PR #54, opened 2011-10-09
Add vertical-align support for images

andreasisaak:
Please add the vertical-align to prevent the 1px space at the bottom of images


Yeswanth-JG:
parent-selector {
 font-size:0;
}

parent-selector * {
font-size:12px (or some desired value)
}

Try the above for removing space in bottom. No need to change default values.

> By default, image will have `"display:inline-block"` and `vertical-align:baseline`

PR #164, opened 2013-02-26
[hidden] attribute should have high precedence

zzzzBov:
the `[hidden]` attribute should take precedence over other `display` declarations, otherwise conflicting information could be presented to assistive technologies. It wouldn't make sense for a screen reader to ignore an element marked as "hidden" while it's clearly visible on the page.

If the element is to be shown, the `[hidden]` attribute should be removed, not overridden.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -48,7 +48,7 @@ audio:not([controls]) {
  */
 
 [hidden] {
-    display: none;
+    display: none !important;
 }
 
 /* ==========================================================================


fritx:
+1 for important..

battaglr:
@fritx, we need to try to match the browsers behaviour as much as possible, and adding `!important` would contradict that. I haven't tested it recently, but I assume that it haven't changed, right?

thierryk:
The interesting thing here is that an element with the attribute `hidden` *should not be made visible*. I believe that should be everyone's expectation. So could we add `!important` here for the same reason we set `margin:0` on `body` or is that meant to be the unique exception to normalize?
For what it is worth, I'm against adding exception, but I just wanted to throw that in...

battaglr:
@thierryk, in a note the [spec for `hidden`](https://www.w3.org/TR/html5/editing.html#the-hidden-attribute) says:

> Because this attribute is typically implemented using CSS, it's also possible to override it using CSS. For instance, a rule that applies 'display: block' to all elements will cancel the effects of the hidden attribute.

Also [suggest that browsers could implement this using this —simplified— rule](https://www.w3.org/TR/html5/rendering.html#hiddenCSS):

```css
[hidden]  {
  display: none;
}
```

In my opinion that's a very bad decision made in the spec, and browsers are just following it. [Apparently other people are not happy about this either](http://lists.w3.org/Archives/Public/public-whatwg-archive/2014Apr/0094.html) —and it's a really interesting conversation.

Anyhow, while adding `!important` will fix the specs *error*, we will end up adding a behaviour that is not present in any browser. And I'm not sure if that is something we should do.

PR #180, opened 2013-04-06
Remove hgroup tag (removed from html5)

malukenho:
The tag <hgroup> has removed from html5 especification


LasseRafn:
Please remove this.. It's 2 years now, and should not be used on any sites that also uses Normalize.

PR #269, opened 2014-01-08
Remove intrinsic min-width of fieldset

vanthome:
At least firefox [1] and chrome [2] have a legacy/ intrinsic min-width
for the `fieldset` element. 
This removes it. The fix is especially useful on mobile.

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=504622
[2] http://stackoverflow.com/questions/8084343/google-chrome-fieldset-overflow-bug

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -264,6 +264,7 @@ fieldset {
     border: 1px solid #c0c0c0;
     margin: 0 2px;
     padding: 0.35em 0.625em 0.75em;
+    min-width: 0;
 }
 
 /**


crismanNoble:
@vanthome Why did you decide to close this? I was going to submit a PR for the exact same thing.

vanthome:
@chrismanNoble  that is an excellent question. I though about it and tried to figure out but I can't. The only thing I can remember is that I changed my mind that this might not be a good idea to have it in normalize.css.

crismanNoble:
@vanthome Do you still feel that way? It seems to me it fits within the normalize.css philosophy. I'd be happy to open a new PR if @necolas or the team thinks it would be a good fit.

vanthome:
@crismanNoble I only remember that I had concerns regarding this because of the consequences on mobile devices. You might want to check the outcome on them.
If nothing breaks, I would again opt for having this.

PR #342, opened 2014-06-11
Font-weight:bolder for fonts with multiple weights

interpaul:
Improvement for fonts that have multiple weights.  
If some block, for example header, is bold and has `strong` inside we can make `strong` bolder if font has additional weights.
 
```html
<h1>Header with <strong>nested <strong>strong</strong> inside</strong></h1>
```

Check this [demo](http://cssdeck.com/labs/x20utix0).

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -110,11 +110,18 @@ abbr[title] {
 
 /**
  * Address style set to `bolder` in Firefox 4+, Safari, and Chrome.
+ * 1. Take font-weigh value from parent element
+ * 2. Set font-weight bolder that parent 
  */
 
 b,
 strong {
-  font-weight: bold;
+  font-weight: inherit; /* 1 */
+}
+
+b,
+strong {
+  font-weight: bolder; /* 2 */
 }
 
 /**


necolas:
What is the last version of Safari that has this bug?

interpaul:
Safari 7.0.4 (Mac) had this bug.

jonathantneal:
This is a great idea because it respects the intent in the default styling of `<strong>`. I will check the current behavior in Firefox to see whether this is still the expected behavior in any UA. I will also check for the inherit bug in Safari 7.0.4.

If no UA observes this behavior, and especially if they have moved away from this behavior, then the change would not be a true normalization. While I think it is an improvement, and Normalize contains [some](https://github.com/necolas/normalize.css/blob/master/normalize.css#L406-L421) improvements, this at least needs to be considered.

jonathantneal:
The current Firefox 43 release styles `<strong>` as `font-weight: bolder`, which makes this a normalization choice between UAs. Since this alters [existing behavior](https://github.com/necolas/normalize.css/blob/master/normalize.css#L113-L120), I’d recommend this make the next minor version tick release. Speaking of which, the current comment reads *Address style set to `bolder` in Firefox 4+, Safari, and Chrome*, but this is not the current behavior in Safari or Chrome.

I also tested for the double-applied bolder bug in all current browsers as well as Safari 6.0.5 (Lion), Safari 6.2 (Mountain Lion) and Safari 7.1 (Mavericks) following the [bug report](https://bugs.webkit.org/show_bug.cgi?id=20720) and the [provided test](http://cssdeck.com/labs/ppukiaa7). Only Safari 6.0.5 (Lion) exhibited the issue. Setting `font-weight: inherit; font-weight: bold;` within the same role did not correct the issue, but creating separate rules did correct the issue.

You can test for the double-applied bolder bug [here](http://s.codepen.io/jonneal/debug/qbpzZo) and test this solution yourself:

```css
strong {
	font-weight: inherit;
}

strong {
	font-weight: bolder;
}
```

Be warned, a good minifier may destroy this necessary repetition.

jonathantneal:
This has been merged. Thanks, @interpaul!

PR #356, opened 2014-07-31
`-webkit-appearance: textfield;` not work in iOS7

sofish:
change `textfield` to `none` make it works on all devices


necolas:
Please can you add a bit more context? What's the issue you're experiencing and what are the consequences of changing the value – doesn't it remove lots of other styles?

jonathantneal:
@sofish, would you be able to provide more context as we asked? If not, I’ll be closing this request at the end of the week.

PR #381, opened 2014-10-31
Combined duplicate style declarations

coltonfitzgerald:
font-weight: bold and overflow: auto had 2 style declarations through the document. By combining these, we can make normalize.css smaller :)


necolas:
easier not to group unrelated tags, even if they share the same declarations. makes no difference to the output either.

PR #384, opened 2014-11-13
Added missing main property

srn:
This is useful if you want to require normalize.css from fx. [webpack](http://webpack.github.io/).


necolas:
thanks, merged

PR #387, opened 2014-12-01
Normalize `hr` overflow

jnowland:
As you can put pseudo elements on hr elements the overflow property is now something that should be 'normalized'.

This property in IE7 was set to visible which is the same as what chrome and firefox currently do.

However IE 8, 9, 10 set overflow to hidden.

Usecase: http://codepen.io/jnowland/pen/JoYKNP

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -205,13 +205,15 @@ figure {
 }
 
 /**
- * Address differences between Firefox and other browsers.
+ * 1. Address differences between Firefox and other browsers.
+ * 2. Address `overflow` set to `hidden` in IE 8/9/10/11.
  */
 
 hr {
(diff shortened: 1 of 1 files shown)


jonathantneal:
I have confirmed that overflow is `visible` in Chrome 16+, Safari 5+, and Firefox 4+, and that overflow is `hidden` in Edge 12 and Internet Explorer ≤ 11. Whether one intends to use pseudo-elements or not, this does normalize their display, and helps those using a technique featured in CSS Tricks for at least 5 years.

jonathantneal:
Thanks, @jnowland. Merged as https://github.com/necolas/normalize.css/commit/46e3e0554c7b81fbfa5cfa1d1531e63f860bb741

PR #392, opened 2014-12-13
Add main field to package.json

jhnns:
This is especially useful when normalize.css is bundled with webpack, which resolves import statements like node.


necolas:
done

PR #394, opened 2014-12-24
Removed obsolete <hgroup> tag

Ftornik:
&lt;hgroup&gt; tag was removed from W3C html5 specification in 2013.


necolas:
leaving it in for legacy reasons atm

PR #400, opened 2014-12-29
remove unnecessary vendor prefixes

marcobiedermann:
according to CanIUse you do not need vendor prefixes for `box-sizing` any more.
See: http://caniuse.com/#search=box-sizing

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -209,7 +209,6 @@ figure {
  */
 
 hr {
-  -moz-box-sizing: content-box;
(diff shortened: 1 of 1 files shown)


necolas:
ESR is at 31, so I'll remove the moz prefixes too.

mzgol:
What mobile browsers does normalize.css support? Dropping `-webkit-box-sizing` is fine as long as you don't support Android 2.3.

Issue #357, opened 2014-07-31
Keep outline on links hovered or focused


lespacedunmatin:
Hello !

On links hovered ou focused, you remove the outline (line 97). This harms the accessibility of the website : http://www.outlinenone.com/.

Could you please remove this rule ?

Thanks

nickautomatic:
As far as I can see, this is definitely an accessibility issue. Removing focus styles by default means that unless people know to replace them, any site built on normalize.css as it currently stands will be difficult to navigate with a keyboard.

Your stated aim is to preserve useful browser defaults, and having a visual indication that an element is focused is definitely a useful browser default, because it makes the page keyboard navigable. That's an important accessibility consideration.

For more on the accessibility implications of removing focus styles, see eg. http://webaim.org/blog/plague-of-outline-0/
http://www.456bereastreet.com/archive/200905/do_not_remove_the_outline_from_links_and_form_controls/

FagnerMartinsBrack:
Nobody is removing focus styles man, read the issue.

nickautomatic:
Ah, sorry about that. You're totally right. I came here because I was working on making a project based on normalize keyboard-accessible, and found all focus styles missing, and - too hastily, I guess - jumped to the conclusion that normalize was to blame.

For what it's worth, my confusion was due to the comment in normalize ("Improve readability when focused and also mouse hovered in all browsers."), which set my expectation that I was looking at focus styles (I guess often you see what you're expecting to see...) Changing it to something like "when active and hovered" would avoid any (excuse for) confusion... :)

PR #406, opened 2015-01-21
This comment seems to be causing confusion

FagnerMartinsBrack:
Just got a notification from this old issue: https://github.com/necolas/normalize.css/issues/357

It seems the comment confuses the reader to think what is changed is the `:focus` pseudo-selector when actually it's the active focused state of the element (or just the "active" state).

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -91,7 +91,7 @@ a {
 }
 
 /**
- * Improve readability when focused and also mouse hovered in all browsers.
+ * Improve readability when active focused and also mouse hovered in all browsers.
  */
 
 a:active,


djch:
Seems to me like that comment could use some commas, hyphens or both.

nickautomatic:
Ah, sorry - just saw this after I commented on the previous thread. For what it's worth, I'd take out the word "focused" altogether. Calling the active state "active focused" is still likely to lead to the same confusion, in my opinion.

necolas:
thanks, fixed and will push soon

PR #410, opened 2015-01-25
Create normalize.min.css

joedanhol:
Faster loading times


TrySound:
I think you need to add builder with minificator. I use gulp, it's pretty and simple.

necolas:
I suggest that you use a build tool to optimize your assets. There are several good Node.js tools for doing that with CSS.

samayo:
Why not include a minified version here? like foundation, bootstrap, semantic-ui ... Not everyone uses gulp or front-end package management when developing.

TrySound:
Uglify CLI?

samayo:
Nope. I am just a PHP dev. I threw in normalize to create a simple page. Anyway, I googled and found a minifying site..

PR #415, opened 2015-02-12
Replace duplicate monospace, monospace property with monospace, serif

krisbulman:
serif was incorrectly replaced with a second monospace property in commit 3fe0df0

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -230,7 +230,7 @@ code,
 kbd,
 pre,
 samp {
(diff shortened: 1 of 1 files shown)


Florian-R:
[NOPE](https://github.com/necolas/normalize.css/search?q=monospace&type=Issues&utf8=%E2%9C%93)

krisbulman:
dup

PR #422, opened 2015-03-11
scaling HTML element to window size

anEffingChamp:
In common browsers `html` may not completely fill the viewport unless otherwise specified. Specifically  `html` will become smaller than the viewport if the viewport is longer than it is wide. Normally this is not noticeable, but it becomes an issue when binding a background image to `html` without explicitly stating `height` and `width`.

```css
html {
  background: url( 'http://upload.wikimedia.org/wikipedia/commons/b/b0/Altare_afasta-se_terra_sol_lua.jpg' );
  background-size: cover;
  height: 100%;
  width: 100%;
}
```

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -8,6 +8,8 @@
 
 html {
   font-family: sans-serif; /* 1 */
+	height: 100%;
+	width: 100%;
   -ms-text-size-adjust: 100%; /* 2 */
   -webkit-text-size-adjust: 100%; /* 2 */
 }


marcobiedermann:
:-1: 
Which browser are you using? I've tried this in latest Chrome and Firefox and it works fine. No need to set height and width. Do you have a demo?

anEffingChamp:
You got it.
http://codepen.io/anon/pen/emPWwE
Resize the window such that the viewport is taller than it is wide, and let the games begin. I tested in Chrome, Safari, and Firefox.

marcobiedermann:
@anEffingChamp Thanks! Yes you are right. You need to set `height: 100%` to fill the entire screen. `width: 100%` has no effect on my end.

necolas:
I don't think normalize.css is the place for this. Your application code is more suitable.

Issue #430, opened 2015-03-30
Include LICENSE.MD when publishing on NPM


arschmitz:
Currently it looks like your publishing from a detached head to NPM and including only package.json, normalize.css, and README.md .  We intend to use this on some jquery projects and currently we technically cant follow the license when using NPM to download the file. The license file is not included, and we plan to commit this file, meaning this would count as distribution with with out the license, and would violate the current license.

Also right now the npm package itself is violating the current license in the repo, as it is distributing source with out the license being included.

PR #433, opened 2015-04-02
Modify license extension

battaglr:
A `.txt` file extension will allow more text editors to know how to open it.

---

- Somehow related to #430.
- H5BP did the same in h5bp/html5-boilerplate#1680.


jonathantneal:
@alrra, you something similar this in H5BP. I know there was [more to the issue](https://github.com/h5bp/html5-boilerplate/issues/1655#issuecomment-78089768) than MD vs TXT, so would you recommend it for other projects?

jonathantneal:
@alrra mentioned that they renamed `LICENSE.md` to `LICENSE.txt` because H5BP is frequently distributed as a ZIP, and in those situations people would not be able to open the license.

battaglr:
@jonathantneal, oh, I see. My reasoning behind this is that since the license is just plain text, the most suitable extension would be `.txt`. Don't hesitate to close this issue if you prefer to maintain de `md` extension; it's just a suggestion. :smiley:

jonathantneal:
Given the reasoning behind the rename, I’ll close this PR. I just wanted to be extra clear and careful. Thanks for all you are doing!

battaglr:
:wink:

PR #437, opened 2015-04-08
Remove hgroup as it is no longer in HTML5

HronakNahar:
See this - http://html5doctor.com/the-hgroup-element/


Florian-R:
[Nope](https://github.com/necolas/normalize.css/search?q=hgroup&type=Issues&utf8=%E2%9C%93)

battaglr:
Also, see [this comment](https://github.com/necolas/normalize.css/pull/180#issuecomment-16005485) in particular.

HronakNahar:
Alright, I understand that we should leave it for legacy reasons. But if we don't remove it (in the later versions), people will keep on using it.

Developers/designers who care (even little) about the web will identify the issue with their hgroup and fix it. And anyway, replacing a tag isn't so hard.

Let's do this: People using the "old" hgroup tag should use the "old" version of normalize.css.

Effeilo:
Some elements have been removed recently, for new projects standards compliant, replace the current rule by:

```css
/**
 * Correct `block` display not defined for any HTML5 element in IE 8/9.
 * Correct `block` display not defined for `main` in IE 11.
 */

article,
aside,
figcaption,
figure,
footer,
header,
main,
nav,
section {
  display: block;
}
```

arthurvr:
I'm -1 on removing `hgroup` although it's something coming up frequently so a note in a comment can't hurt.

battaglr:
@arthurvr: it's a great suggestion to add a comment as a `NOTE` tag —similar to [#L402](https://github.com/necolas/normalize.css/blob/master/normalize.css#L402)— explaining which elements are deprecated by the W3C. Maybe we should open another issue about this?

PR #440, opened 2015-04-23
Update normalize.css

vanesyan:
`hgroup` tag was deprecated about 2 year ago.
Proof: https://www.w3.org/wiki/HTML/Elements/hgroup


alrra:
@vanesyan See [this](https://github.com/necolas/normalize.css/search?q=hgroup&type=Issues&utf8=%E2%9C%93). :)

vanesyan:
Sorry, my bad. I should had to look for the same, before sending pr.

PR #443, opened 2015-04-30
very small change

starlingDesign:
line 185, changed "img" selector to "a img" so  it only applies to images inside <a> as that is the target fix.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -182,7 +182,7 @@ sub {
  * Remove border when inside `a` element in IE 8/9/10.
  */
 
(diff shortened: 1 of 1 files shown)


henrik:
This also increases the specificity of the selector, though.

Did the current selector cause any issues for you, that this change fixes?

starlingDesign:
not so much an issue, but if I do specify a border on a photo, it simply has to "overwrite" the normalize rule.  I guess there is no harm in keeping it more generalized.

PR #452, opened 2015-05-19
More specific rule for removing border of an <img> inside an <a>

vlakoff:
Fixes this use case:

    .someimg {
        border-width: 2px;
        border-color: green;
    }

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -183,7 +183,7 @@ sub {
  */
 
 img {
-  border: 0;
+  border-width: 0;
 }
 
 /**


uniZero:
The example you've provided doesn't refer to an  inside an <a> element. Is this what you mean?

```
a img.someimg {
    border-width: 2px;
    border-color: green;
}
```

vlakoff:
Indeed, the `img` are inside `a`. I encountered the issue transcluding some third-party content. But I just noticed I had overlooked that `border` attribute:

    <a ...></a>
    + CSS like the above

It is as shitty as the rest of the code they're producing.

I was doing some tests with `border-style:none`, but let's just keep the reset CSS as it is, the `border:0` is largely used and fine.

vlakoff:
I'm closing it, as there is nothing to do ;-)

PR #465, opened 2015-07-13
Fix <sup> property

lnfnunes:
<sup> text isn't aligned with normal text cropping the top a little bit.
Check it out: http://codepen.io/lnfnunes/pen/vOrKeV

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -168,7 +168,7 @@ sup {
 }
 
 sup {
-  top: -0.5em;
+  top: -0.25em;
 }
 
 sub {


robsonsobral:
Which browser?

Tigt:
Screenshot of the text being cut off as described

Happening to me in Firefox for OS X.

lnfnunes:
- Chrome 43 for (Linux / Windows)
Chrome

- Firefox 39 (Windows)
Firefox

- Internet Explorer 11 (Windows)
IE

battaglr:
Also happening on Chrome 43 for OS X.

@lnfnunes: `.25` seems a little low; maybe we can try with a `.3`, `.35` or `.4` —we also need to consider the font in use.

lnfnunes:
Looks like the major render problem is in Chrome and  I´ve realized setting the _line-height_ to &lt;sup&gt; parent element fixed this issue even with font-family variations.
```css
p { line-height:1.2 }
```
I´ve updated the test case: http://codepen.io/lnfnunes/pen/vOrKeV

I don´t think be possible to normalize.css fix this... __unless__ adding the line-height to all "text" elements.
Does it make sense?

PR #490, opened 2015-12-01
Replace lost button and input focus style

### LINKED ISSUE #481: Removing button border removes focus outline in FF
It's fairly common to want to remove button borders. If you do this while using normalize, the :focus style that keyboard users depend on will disappear in Firefox.

Tab through the links and buttons in the following examples for a demonstration:
* With normalize http://codepen.io/anon/pen/Gpmmqw/
* Without normalize http://codepen.io/anon/pen/rOmmKz

This is due to https://github.com/necolas/normalize.css/blob/master/normalize.css#L313

I'm using Firefox 40.0.3 on OS X

tdhooper:
Fix #481

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -315,6 +315,15 @@ input::-moz-focus-inner {
 }
 
 /**
+ * Replace focus style removed in the border reset above
+ */
+
+button:-moz-focusring,
+input:-moz-focusring {
+  outline: 1px dotted ButtonText;
+}
+
+/**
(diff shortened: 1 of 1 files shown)


jonathantneal:
Great work, @tdhooper!

mgol:
@jonathantneal Could you tag a new version & publish to npm? I was just hit by #481 and found that it's already fixed here and just not released yet.

Issue #471, opened 2015-08-28
box-sizing issue with input[type="search"]


patrickhlauke:
```
-webkit-box-sizing: content-box;
box-sizing: content-box;
```

appears to be making search inputs wider than they should be - see https://github.com/twbs/bootstrap/issues/17379

BYODKM:
#358

patrickhlauke:
So, maybe @necolas can confirm (as I just read through the linked issues there, with their very terse comments): is the fact that `box-sizing: content-box` results in search inputs being larger than their equivalent text inputs intentional then, and won't be fixed? /cc @cvrebert

BYODKM:
Normalize.css is for normalize browsers' default style. Which means you can not always get what you want. You can have a small reset css for `input[type=search]` in your own project as usual.

patrickhlauke:
How about normalizing it to what should be the default, according to https://www.w3.org/Bugs/Public/show_bug.cgi?id=28784 ? I'm fine overriding it myself, but effectively here you're normalizing it against all current consistent (with exception of Firefox, but that's being fixed now too) default styling in browsers...

FagnerMartinsBrack:
I am watching this repository for a long time, and as far as I know there was no discussion regarding search inputs being wider than they should be. Most discussions regarding that style is about it being harder to override, because one can't just include normalize and go for `* { box-sizing: border-box; }` and expect the search inputs to be overriden.

Other than that this seems pretty legit and worth looking into, as long as it doesn't cause a regression from the [first bullet point](https://github.com/necolas/normalize.css/blob/3.0.3/normalize.css#L352).

patrickhlauke:
As there seems to be little interest here in addressing this, closing. We fixed it back to browser defaults in Bootstrap, FWIW.

necolas:
Sorry for being absent. I agree that this is an annoying quirk of normalize that should be fixed. It looks like the original reasoning doesn't stand up, and I'd rather be pragmatic about it anyway.

patrickhlauke:
Sorry for interpreting silence for lack of interest/agreement. Wasn't sure how best to formulate a visual test, but gave it a go in the PR anyway.

PR #496, opened 2016-01-05
Remove explicit box-sizing for input[type="search"]

patrickhlauke:
This normalizes the CSS to what the suggested default browser CSS should be, which is now implemented in IE11, Edge, Blink, WebKit and Gecko (for the latter, see https://www.w3.org/Bugs/Public/show_bug.cgi?id=28784)

References https://github.com/twbs/bootstrap/issues/17379

Closes https://github.com/necolas/normalize.css/issues/471

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -348,13 +348,11 @@ input[type="number"]::-webkit-outer-spin-button {
 }
 
 /**
- * 1. Address `appearance` set to `searchfield` in Safari and Chrome.
- * 2. Address `box-sizing` set to `border-box` in Safari and Chrome.
+ * Address `appearance` set to `searchfield` in Safari and Chrome.
  */
 
 input[type="search"] {
-  -webkit-appearance: textfield; /* 1 */
-  box-sizing: content-box; /* 2 */
+  -webkit-appearance: textfield;
 }
 
 /**
--- a/test.html
+++ b/test.html
@@ -429,12 +429,6 @@
   <div class="Test-run">
     <input type="search" style="border:1px solid #ADD8E6; padding:10px; width:200px;">
   </div>
-  <h3 class="Test-it">should have a <code>content-box</code> box model</h3>
-  <div class="Test-run">
-    <div style="background:red; display:inline-block; height:62px; width:242px;">
-      <input type="search" style="border:1px solid #ADD8E6; height:20px; padding:20px; width:200px;">
-    </div>
-  </div>
   <h3 class="Test-it">should not have a cancel button in Safari or Chrome</h3>
   <div class="Test-run">
     <input type="search" value="search">


normalize.css
```diff
@@ -349,12 +349,12 @@ input[type="number"]::-webkit-outer-spin-button {
...
 /**
  * 1. Address `appearance` set to `searchfield` in Safari and Chrome.
- * 2. Address `box-sizing` set to `border-box` in Safari and Chrome.
+ * 2. Address `box-sizing` set to `content-box` in old versions of Firefox
  */
 
 input[type="search"] {
   -webkit-appearance: textfield; /* 1 */
-  box-sizing: content-box; /* 2 */
+  box-sizing: border-box; /* 2 */
```
csshugs: Why not remove the `box-sizing` declaration entirely and leave the line blank?

When you need it to be `border-box`, you likely [set everything to border-box](http://www.paulirish.com/2012/box-sizing-border-box-ftw/). And if anyone really should just need the <input type="search"> being `border-box`, I think he is better on doing that in his own styles.

patrickhlauke: to normalize older versions of Firefox (and potentially other, less common browsers with non-standard defaults)? otherwise fine, could also just be removed altogether

necolas: What about setting border-box on `input`? I noticed that `input[type=number]` seems to use `content-box` in Chrome. And this would also keep the specificity at 0,0,0,1 which is easy to override; I think the specificity of this rule is probably more annoying/problematic than the styles it applies.

patrickhlauke: interestingly, despite what is said here https://html.spec.whatwg.org/multipage/rendering.html#form-controls it seems most browser do indeed set `box-sizing:content-box` in their browser styles. wonder if now forcing this sizing generally would end up breaking layouts for devs who relied on this. perhaps the least harmful path would indeed be to just drop the `box-sizing` rule altogether from `input[type="search"]`, now that firefox has harmonised with other browsers ?

csshugs: > perhaps the least harmful path would indeed be to just drop the box-sizing rule altogether from input[type="search"]

Agree.

> What about setting border-box on input

Wouldn't that be a minor update rather than a patch? I'm afraid it would break peoples styles in some cases, doesn't it?

patrickhlauke:
(btw happy to squash / rebase this PR ... just didn't want to jump the gun here in case the decision to just nuke the `box-sizing` wasn't final)

patrickhlauke:
Rebased this to just remove the box-sizing CSS and test, as discussed.

necolas:
Thanks. Applied as fe567632afcf8afcf17925ab2a507ed87504cce8

PR #497, opened 2016-01-07
IE8 keeps elements hidden, when el.hidden = false;

thisconnect:


cvrebert:
Strange, because `hidden` is a [boolean attribute](https://html.spec.whatwg.org/multipage/infrastructure.html#boolean-attribute), so `hidden` and `hidden="false"` ought to be equivalent (and result in hiddenness), and `hidden="false"` is invalid per the spec.
And setting `HTMLElement.hidden = false` ought to just remove the attribute altogether.
But this is IE8 we're talking about. I'll perform a test.

cvrebert:
So, the IDL attribute indeed behaves incorrectly, but that's technically more of a DOM bug. Not sure whether it's worth addressing at the CSS layer.

Testcase: http://jsbin.com/zusuto/edit?html,css,js,output
Screenshot:
ie8

jonathantneal:
The values "true" and "false" are not allowed on boolean attributes. To represent a false value, the attribute has to be omitted altogether, so this is definitely a DOM bug. @thisconnect, would you be interested filing this issue with [polyfill-service](https://github.com/Financial-Times/polyfill-service/issues/new)? The fix would be JS. Maybe something like:

```js
Object.defineProperty(Element.prototype, 'hidden', {
	get: function () {
		return this.hasAttribute('hidden');
	},
	set: function (value) {
		if (value) {
			this.setAttribute('hidden', '');
		} else {
			this.removeAttribute('hidden');
		}
	}
});
```

thisconnect:
quote:
> Looking at @jonathantneal's comment here I'm wondering why this is a bug. If boolean attributes are not allowed to have values, and so a value of false is equivalent to true (because it's the presence of the key rather than anything to do with the value), hiding an element when it has a hidden property would be correct behaviour, no?

PR #503, opened 2016-01-23
Add 3.0.3

battaglr:
Add v3.0.3. This also updates all website references and latest file, which is used in the download link placed in the docs.


battaglr:
This resolves #489.

jonathantneal:
@battaglr, does this v3.0.3 version include some of the larger pulls we’ve accepted the last week? If so, how do you feel about holding these for v4? My concern is that some are breaking changes.

These larger changes include removing box-sizing on input[type="search"], and then removing color inheritance for button, input, optgroup, select, and text. Thoughts?

battaglr:
Oops! My bad! I took the code from *master* and not from the [v3.0.3 tag](https://github.com/necolas/normalize.css/releases/tag/3.0.3). I'll fix that soon.

What do you think about creating a "development" branch —with the name you prefer? Having changes on *master* that are not part of any release yet could be confusing, and the people that consume the CSS file directly through GitHub potentially would experience problems, since as today, we have braking changes in *master*.

Finally, I totally agree that the recent changes must be added in a new mayor version.

battaglr:
@jonathantneal, please, check this again; this time should be right. :sweat_smile:

jonathantneal:
Looks good, but I’ll look again after my morning to be sure I miss anything. Thank you, @battaglr.

jonathantneal:
Thanks!

battaglr:
:+1:

PR #487, opened 2015-11-05
Adding license info to bower.json

reiz:

PR #504, opened 2016-01-23
Remove Bower and Component installation

battaglr:
Bower and Component installation support was dropped in 8cfa3e765e0cb55c174657e412430424a9fee97a.

### DIFF
--- a/index.html
+++ b/index.html
@@ -31,8 +31,6 @@
 
         <div class="txt-large">
           <pre><code><a href="https://www.npmjs.org/package/normalize.css/">npm</a> install normalize.css</code></pre>
-          <pre><code><a href="https://component.github.io/">component</a> install necolas/normalize.css</code></pre>
-          <pre><code><a href="http://bower.io/">bower</a> install normalize.css</code></pre>
         </div>
 
         <div class="share-bar">


jonathantneal:
Thanks, @battaglr! You have been so helpful.

battaglr:
Happy to help!

Jmeyering:
Whoa...why the hell was bower dropped?

t1st3:
@Jmeyering see #455 , #487 or #464.

Anyway, you can still get this package through bower like this:
```shell
bower install git://github.com/necolas/normalize.css.git#master
```
and replace `master` with the version you want to get.

Jmeyering:
Sad day! Bower is very nice for frontend assets. Count me as one of the backward individuals who don't think npm is very appropriate for frontend dependencies. I know I know, that's so old school :)

jonathantneal:
@Jmeyering, did the alternative syntax that @t1st3 suggested allow you to keep using normalize with bower?

t1st3:
@jonathantneal I tested it, with the following command, and it installs v3.0.3 smoothly:
```shell
bower install git://github.com/necolas/normalize.css.git#3.0.3
```

Note that one could also perform the same install with writing the following in `bower.json` (also tested a few minutes ago):
```json
{
  "name": "my-package",
  "dependencies": {
    "normalize.css": "necolas/normalize.css#3.0.3"
  }
}
```

So basically, both the CLI-way and the `bower.json`-way work fine. One just has to a bit more precise in their notation using either way.

Since Bower relies on Git tags, both will work, even if support for Bower has been removed. Actually, you can bower any git repo, even if it has not been published on Bower...

Jmeyering:
Right it works fine...I'm more so griping against the entire trend toward npmifying everything...But I'm behind the curve it looks like. No more need to bother you guys, thanks for the input!

tjwudi:
I would say this is a breaking change and should be included in v4 instead of v3 since it could break things. There are many information we utilize in **bower.json** like `main` files.

elantion:
I never think NPM is good for browser. I know NPM and Bower both manage Javascript packages. But for my opinion, Bower is much better play well with browser. 
These day I struggle by using Browserify, Webpack, Jspm and so on. After learning ton of things I still come back to Bower and RequireJs. I still don't understand why they make modules so complicate.
I really don't know why people keep trying to merge Node.js and browser environment. They both written in JavaScript but it doesn't mean they should go together.
Sorry out of this topic.

battaglr:
@tjwudi, hi! Can you elaborate on why you consider this a breaking change? Thanks!

tjwudi:
@battaglr Some tools like [wiredep](https://github.com/taptapship/wiredep) depends on information extracted from **bower.json** and is now in many build environments. Simply removing **bower.json** will absolutely break them.

battaglr:
Bower and Component support was removed on 8cfa3e765e0cb55c174657e412430424a9fee97a and discussed on #455.

Since I don't use these kind of tools —or Bower for some time now— I'm not sure what is the best step to take.

If this is as breaking as some of you are saying, maybe we could add it again, label it as deprecated —possibly on the *changelog*— and remove it effectively on v4... but I'm just thinking out loud here.

@jonathantneal, @necolas, is this something to be worried about, or is it okay as it is?

tjwudi:
I don't know about component as well, since it has smaller usage I believe. But bower, it could be an issue to remove it.

dcdeiv:
Please, add support to bower again! :sob: 
Now I have to install **normalize.css** as a **git submodule** because I have to ignore the node_modules folder. I hate installing dependencies via npm, I use npm to install devDependencies! I prefer installing dependencies via Bower.

Bower makes life easier, please, don't make things complicated!!

tjwudi:
@dcdeiv ++

TrySound:
@dcdeiv What's wrong with bower install repo?

dcdeiv:
@TrySound what's wrong in officially support bower?

TrySound:
@dcdeiv Forcing people to stop using bower :)

dcdeiv:
@TrySound so you want to stop people from using bower? lol

tjwudi:
@TrySound What's wrong with copy-paste normalize.css and why bother install it?

dcdeiv:
@tjwudi I guess they just don't care about delivering their product as people want. Like they force people taking a ship instead of an airplane!

elantion:
Browserify is a good tool to make them(server js and browser js) together.But I am still confusing how to use it in desktop website. It seem only good fit for web app.

TrySound:
@tjwudi You can still use repo with bower. @dcdeiv bower is not airplane. I was bower lover but it's really not the best decision. It has many problems from the start (for example multiple `main`).

dcdeiv:
You have just one main so no problem, isn't it?

TrySound:
@dcdeiv I mean `main` as array. It's part of the spec.

elantion:
@TrySound You can read this article: http://blog.npmjs.org/post/101775448305/npm-and-front-end-packaging
Npm still don't implement browser side feature. They just have a plan.

TrySound:
@elantion Of cause file system (or similar) based module won't work in browser. I don't think it's problem. I use rollup and postcss-import to build my assets. They works perfectly. And yes rollup requires the same and files with es2015-modules for smaller bundle. npm do everything good. The problem can be only node specific. css projects can use `style` field. Assets can be rebased via something like postcss-assets or postcss-copy.

elantion:
@TrySound Yes, I know some kind of tools to get rip of bower. For example, webpack, jspm, browserify etc. I try these tools before, They are play very well with web app. But in traditional desktop website I have to write a lot of configuration yet can not play well. Maybe I need to dig more information. I think requirejs and bower is still a good combination even it has a bit more http request.

dcdeiv:
Bower is immediate! You don't have to learn anything to use it!

TrySound:
@dcdeiv You need to learn `bower`.

dcdeiv:
You are delivering a simple css file. No special config needed! :)

TrySound:
@dcdeiv For your case `bower install repo` is all you need. Other cases require tools which you don't want to learn.

dcdeiv:
Just simpler typing normalize.css than https://github.com/whoever/normalice.css#version

TrySound:
@dcdeiv `bower i necolas/normalize.css`. Okay with name a bit more, but still fine.

tjwudi:
@TrySound Regardless of keeping **bower.json** or not, please confirm that this change could break application's building system and should be included in next major version instead.

TrySound:
@tjwudi Sure. It will.

Issue #508, opened 2016-02-05
Use Stylelint as a gatekeeper


jonathantneal:
[Stylelint] is a modern CSS linter that helps enforce consistent conventions and avoid errors in stylesheets. A `.stylelintrc` file could be used to enforce this project’s coding conventions, and serve as a pull request gatekeeper.

[Stylelint]: https://github.com/stylelint/stylelint

jonathantneal:
That was fast. Thanks, @hudochenkov!

PR #511, opened 2016-02-05
Use Stylelint as a gatekeeper #508

hudochenkov:
I've configured stylelint for this project specifically. No need to have stylelint installed globally because npm using stylelint installed locally (after `npm install`).

I've used `test` for npm script name for easy adding Travis CI (#512).


jonathantneal:
Wonderful! This has been merged.

PR #530, opened 2016-03-15
Use explicit `outline-width` property instead of `outline`

### LINKED ISSUE #524: outline : 0  ▶ outline-style: none;
I found it in your code

```
a:active,
a:hover {
  outline: 0;
}
```

According to reference outline property is 

> outline-color outline-style outline-width|initial|inherit

Therefore, I'd like to do so in order to avoid any confusion.

```
a:active,
a:hover {
  outline-style: none;
}
```

battaglr:
By using `outline-width` we are explicit in our intention.

Also as this is an [opinionated style](https://github.com/necolas/normalize.css/blob/3.0.3/test.html#L153), it's now marked as such.

*This closes #524.*

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -90,12 +90,12 @@ a {
 
 /**
  * Improve readability of focused elements when they are also in an
- * active/hover state.
+ * active/hover state (opinionated).
  */
 
 a:active,
 a:hover {
-  outline: 0;
+  outline-width: 0;
 }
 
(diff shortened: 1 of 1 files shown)


jonathantneal:
Thanks! I tested this against the base supported browsers to be sure `outline-width` worked. https://developer.mozilla.org/en-US/docs/Web/CSS/outline-width#Browser_compatibility

PR #531, opened 2016-03-15
Use inherit to fix legend color

### LINKED ISSUE #363: Fix the inheritance of color for `legend` by the right way
Hi nicolas,

Since you dropped the support for IE 6/7, the legend can inherit the color by this way:
```
legeng {
  color: inherit;
}
```
IMHO, that will be more understandable for everyone than the present hack with `border`.
To be sure, I tested this method with this page: [legend.html](http://www.7studio.fr/github/normalize.css/legend.html)

OS / Browser  | Test 
------------- | ----
WinXP / IE8   | Passing
Win7 / IE9    | Passing
Win8 / IE10   | Passing
Win8.1 / IE11 | Passing

jonathantneal:
Use inherit to fix legend color and clarify that the fix applies to `<legend>` within `<fieldset>`.

This closes #363.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -397,12 +397,12 @@ fieldset {
 }
 
 /**
- * 1. Correct `color` not being inherited in IE 8/9/10/11.
+ * 1. Correct `color` not being inherited from fieldset in IE 8/9/10/11.
  * 2. Remove padding so people aren't caught out if they zero out fieldsets.
  */
 
 legend {
-  border: 0; /* 1 */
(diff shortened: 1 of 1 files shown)


battaglr:
@jonathantneal, looks good, just remember to remove [the "legend" section from the docs](https://github.com/necolas/normalize.css/blob/3.0.3/README.md#legend), since it will no longer apply.

jonathantneal:
Noted. I will also need to update the changelog

PR #535, opened 2016-03-17
Only normalize border-style on img

jonathantneal:
Correct `border-style` given when inside `a` element in IE 8/9/10. See http://s.codepen.io/jonneal/debug/274c26a11e945ee932aa1f32512d48e6

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -188,11 +188,11 @@ sub {
    ========================================================================== */
 
 /**
- * Remove border when inside `a` element in IE 8/9/10.
+ * Correct border-style given when inside `a` element in IE 8/9/10.
  */
 
 img {
-  border: 0;
+  border-style: none;
 }
 
 /**


normalize.css
```diff
@@ -188,11 +188,11 @@ sub {
    ========================================================================== */
 
 /**
- * Remove border when inside `a` element in IE 8/9/10.
+ * Correct border-style given when inside `a` element in IE 8/9/10.
  */
 
 img {
-  border: 0;
```
battaglr: @jonathantneal, although `border-style: 0` may work on some implementations, the proper value should be `none `.

normalize.css
```diff
@@ -188,11 +188,11 @@ sub {
...
+  border-style: 0;
```
battaglr: @jonathantneal, although `border-style: 0` may work on some implementations, the proper value should be `none `.

PR #536, opened 2016-03-18
Correct wrapping not present in IE8/9/10/11 and Edge 12/13

### LINKED ISSUE #100: Fix IE9 legend text wrapping
IE9 doesn't wrap `legend` text.

A quick look for a solution shows that adding `max-width: 100%` will cause IE9 to wrap text.

Need to do some tests to understand the impact of adding that declaration, e.g., people being caught out if they try to style the legend element and modify other box-model properties. Also check to see if there is an alternative fix that is cleaner.

jonathantneal:
Corrects `<legend>` wrapping not present in IE8-11 and Edge 12-13. It actually requires 2 separate fixes in order to work in both IE8 and Edge 13.

`box-sizing: border-box` and `max-width: 100%` correct the issue in IE9-11 as well as Edge 12-13, while `display: table` and `white-space: normal` correct the issue in IE8-11.

Resolves #100

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -388,13 +388,18 @@ fieldset {
 }
 
 /**
- * 1. Correct `color` not being inherited from fieldset in IE 8/9/10/11.
- * 2. Remove padding so people aren't caught out if they zero out fieldsets.
+ * 1. Correct wrapping not present in IE8/9/10/11 and Edge 12/13.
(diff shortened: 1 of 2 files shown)


battaglr:
This code looks technically correct for me! :+1:

PR #537, opened 2016-03-18
Reduce input type selector weight

### LINKED ISSUE #438: Why the extra specific selector `html input[disabled]`?
This:

``` css
html input[disabled] {
  cursor: default;
}
```

means I have to get more specific to override it.

I'm sure there's good reason to have the `html` in there, but I'd love to understand why – and maybe a comment could be added to explain it? Happy to contribute that comment once I understand the reason.

jonathantneal:
- Moves input type appearance normalization into its own rule.
- Reduces selector weight of input types.

Resolves #438

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -293,8 +293,6 @@ select {
  * 1. Avoid the WebKit bug in Android 4.0.* where (2) destroys native `audio`
  *    and `video` controls.
  * 2. Correct inability to style clickable `input` types in iOS.
- * 3. Improve usability and consistency of cursor style between image-type
- *    `input` and others.
  */
 
 button,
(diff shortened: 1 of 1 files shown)


normalize.css
```diff
@@ -293,24 +293,31 @@ select {
...
+  cursor: pointer;
 }
 
 /**
  * Re-set default cursor for disabled elements.
  */
 
-button[disabled],
-html input[disabled] {
+[disabled] {
```
battaglr: @jonathantneal, this rule is now also going to impact on `select` elements. Nothing wrong with that; just pointing this out because it's a new behavior.

battaglr:
1. Placing `appearance ` on it's own rule-set LGTM! (thanks @henrik!).
2. The reduction on selector weight LGTM (I would like to give 1 million :+1: to this)!  The numbers in your tests are a pleasant surprise for me (thanks @benfrain! for your original tests!).

Overall: LGTM :+1:

benfrain:
Glad they helped.

FWIW — I fear (but not checked) that occasionally user agent style sheets add a rule that includes type and attribute (Safari I think) so reducing specificity might actually cause a problem with some inputs as they may not be specific enough to undo the UA style sheet.

jonathantneal:
@benfrain, User Agent styles are unique in that they have no weight!

A less-weighted selector beating a heavier-weighted selector from the User Agent stylesheet

benfrain:
Ah, great! :)

thierryk:
> User Agent styles are unique in that they have no weight!

FWIW, I ran once in an issue where WebKit (its UA styles sheet) was styling an element using `!important` which did make my own styling irrelevant. I can't remember what declaration it was though...

benfrain:
That sounds similar to the issue I remember hitting. But like @thierryk I can't remember the specific selector scenario.

I guess if they are slapping !important on something in the UA style sheet that's a problem best addressed outside of this anyhow.

Issue #55, opened 2011-10-15
code inherits word-wrap: break-word from a parent pre


timmolendijk:
When you have the situation in which code is put in `<pre><code>...</code></pre>` blocks -- a fairly common practice -- the explicitly set `word-wrap: break-word` on `pre` is problematic, as it will be inherited by the `code` element (in which it tends to be very undesirable).

Two options in my opinion:
- remove the explicit `word-wrap: break-word` on `pre`, or;
- add an explicit `word-wrap: normal` to `code`.

PR #538, opened 2016-03-18
Remove opinionated pre overflow

### LINKED ISSUE #313: Problem with pre overflow auto.
Hi,

Not a big problem, but I had to remove the pre { overflow: auto: } specification from my v3.0.0 normalize.css so that my pre tags within a div { overflow: auto; } would scroll properly. The problem seems to be that at least in Chrome (Mozilla/5.0 (X11; Linux i686) AppleWebKit/537.36 (KHTML, like Gecko) Ubuntu Chromium/32.0.1700.107 Chrome/32.0.1700.107 Safari/537.36), you can't undo the overflow: auto; with some other setting. The pre element scrolls horizontally first.

To be specific, I want white-space: pre, and not white-space: pre-wrap; So this is more of a heads-up, than a real issue. Many users might not care too much. Not sure, but by a quick search of past issues, this might be something that may be better set per installation. It seems to be quite fluid as to what is the best "default" value for each application. Just a thought.

jonathantneal:
This removes the opinionated view that `pre` should contain its overflow.

Not necessary to merge unless there is more consensus.

Resolves #313

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -227,14 +227,6 @@ hr {
 }
 
 /**
- * Contain overflow in all browsers.
- */
-
-pre {
-  overflow: auto;
-}
-
-/**
  * 1. Correct inheritance and scaling of font-size for preformatted text.
  * 2. Address odd `em`-unit font size rendering in all browsers.
  */
--- a/test.html
+++ b/test.html
@@ -241,12 +241,6 @@
     <hr style="height:2px; border:solid #ADD8E6; border-width:2px 0;">
   </div>
 
-  <h2 class="Test-describe"><code>pre</code></h2>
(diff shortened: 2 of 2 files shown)


battaglr:
@jonathantneal, LGTM in case you decide to merge this. :+1:

jonathantneal:
This rule has changed a few times over the years. See the [initial commit of normalize.css](https://github.com/necolas/normalize.css/blob/d8335d302adef6fe2632925167c358ef1c7c6145/normalize.css#L110-L119), the concern over that in #55, and later a8edd0c, and the concern over that in #313. :)

I feared removing an opinionated style until I read Nicolas’ response to concern over the previous behavior:

> I don't think it is going to be as uncommon (or disasterous) as you claim.

That’s my stand, too.

Issue #542, opened 2016-03-20
'main' field dropped from package.json with 4.0.0


pekim:
It was added with 947b1ae124cc987848262b6513242c83348a733d, but has been dropped with 3123b8e9f5923fd06e4eeb9e0727b7cc42fd94f6. This means requiring `normalize.css/normalize.css` instead of just `normalize.css`.

Is this an intentional change?

battaglr:
@pekim, hi! It was indeed an intentional change, see #441. Is this causing you any troubles?

viko16:
Oh, I think we should notice this change on README.

battaglr:
@viko16, you're right, we should add that, but to the [CHANGELOG](https://github.com/necolas/normalize.css/blob/master/CHANGELOG.md), not the README. I've opened #551 to do that. Thanks!

maxdow:
It breaks with webpack or it's now needed to add a custom configuration. 

Typical example when you want to extract vendors modules : 

var vendors = Object.keys(require("../package.json").dependencies);

var webpackConfig = {
  entry : {
    app : [path.resolve(__dirname, "../src/app.js")],
    vendors : vendors,
  },
  plugins : [
    new webpack.optimize.CommonsChunkPlugin({
      name :"vendors",
      filename : "vendors.bundle.js"
    }),
...
}

We now need to add a custom alias ( resolve.alias ) property to the correct file.

Is it a definitive change ?

jonathantneal:
Would it be more appropriate to add `main` back to `bower.json`?

pekim:
It's not really a big deal, and easy enough to deal with. It's more that it came as a slight surprise that's all.

jonathantneal:
@pekim, for some time bower.json had been dropped from the project, but a release containing the change had not been made. We brought it back before the v4 release. This is why I’m mentioning restoring the file. @TrySound might have additional advice, as he was involved in our bringing it back.

jonathantneal:
This has been re-added via #564

benoitvallon:
Since you decided to re-add it (which is nice I think), would it be possible to publish a 4.0.1 on npm please? Thanks

PR #551, opened 2016-03-21
List removal of `main` field from `package.json`

battaglr:
- List removal of `main` field from `package.json` in the CHANGELOG file.
- Mark files with code syntax to be consistent with the CONTRIBUTING file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 * Update comments to identify opinionated styles.
 * Update comments to specify browser/versions affected by all changes.
 * Update comments to use one voice.
+* Remove `main` field from `package.json`.
 
 ---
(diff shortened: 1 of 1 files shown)


SimenB:
Why was `main` removed? And if it'll remain that way, how about adding a `style` field instead, like parceliify expects?

https://github.com/rotundasoftware/parcelify

battaglr:
@SimenB, hi! That's something I don't know much about (parceliify). Further discussion about this is taking place on #542; in here your opinion *may* get lost in the issue tracker. :wink:

jonathantneal:
I’d like to restore `main` to resolve #542

battaglr:
Closing because of #564.

PR #556, opened 2016-03-22
Remove optional quotes from CSS attribute selectors

antibland:


battaglr:
@antibland, hi! IMO that kind of optimizations should be performed by minifiers.

antibland:
@battaglr Maybe minifiers *should* do this, but there is no guarantee that all *will*. Also, This repo appears to be something of a starting point for new projects. Omitting the quotes is a style choice and shaves a few extra bytes.

battaglr:
@antibland, I'm sorry if I was a bit blunt.

This is indeed just a matter of style choice. This project uses an [explicit style guideline](https://github.com/stylelint/stylelint-config-standard/blob/master/index.js) and [an implicit one](https://github.com/necolas/idiomatic-css).

That being said, if we were up to do simple things that would save a few bites, we could avoid the last semi-colon in a rule, or avoid leaving spaces between the selector and the bracket. But I think that's the job of a minifier, not our job.

In any case, remember that the size of "author" code rarely matters; but if you were comparing the two versions, would be fair to compare the *gzipped* size:

- With quotes: 2115 bytes.
- Without quotes: 2111 bytes.

IMO in this context 4 bytes are no real difference.

Issue #450, opened 2015-05-13
Normalize.Sass


TheJaredWilcurt:
I was typing this up on my phone for my other issue and it was closef in the process, since I can't reopen it I'm just making a new one.

* * *

Not everyone uses grunt/gulp. Normalize is something beginners will be using to learn. And beginners who learn more advanced stuff are more likely to run into a GUI based pre-processor like Scout, Koala, Prepros, Hammer, Mixture, etc, than they are to set up all the text files by hand to automate the processes those GUIs perform. For those with the know-how already, having more options doesn't hurt. Sass would be producing an identical css file AND minified css file which Normalize isn't even doing now. This is an improvement.

Having to go to a third party to get a Sass version means you ALSO have to come here anyways to make sure it's up to date. Additionally it means that it won't get the same level of attention as it would if it were part of the main repo. If it's from a third party, that means you don't know if their version is reputable. You'd have to manually go line by line to make sure the third party person converted it correctly and at that point it's quicker to just do it by hand anyway. Having a Sass version as part of the repo by default just acts as a time saver so I don't have to manually do it myself each time there is an update.

It's obvious there are people who want this. I don't understand the resistance to this request. It's a developer desired feature, it saves time, and has the benefit of automatically maintaining the 3 most common style files (.sass, .css, .min.css). This is wins all around.

The only reasons not to do this is:

1. The maintainers are not familiar with Sass. (It's super easy to learn, takes about a week or two to get the whole thing.)
2. The maintainers think Normalize is a long-term project that could outlast Sass (a technology that could be replaced by something else in the future). It doesn't matter because the project can be switched to the new thing or back to plain CSS if that's ever the case, but for the foreseeable future at least we can have the benefits.

* * *

There now this has been made public for the open source community to weigh in on. So you can now feel free to passive-aggressively close this ticket by quoting the same mantra you have to the many others making this request.

nicolasmccurdy:
I don't think the main reasons that the contributors don't want to use SASS are related to a learning curve. A SASS file would be another file to maintain and deal with distributing and hosting. You don't need a stylesheet to be in SASS to use it with other SASS styles (also applies to other preprocessors), and there helpful tools ([like this](http://css2sass.herokuapp.com/)) to convert CSS to other preprocessors like SASS.

fleeting:
[...]

Issue #446, opened 2015-05-09
Official Sass Support


TheJaredWilcurt:
I don't know if Normalize has already been ported to Sass/SCSS, but it's not mentioned in the readme.

I did a quick convert for my own project of 3.0.3. Feel free to merge it into the repo.

* https://gist.github.com/TheJaredWilcurt/7594dbd92c3c2e22256e

florianbouvot:
Please see #9, #22, #26, #30, #240

svenjuhnke:
Porting CSS to SCSS? Isn't it upside down?

Vice versa. SCSS as masterfile.

TheJaredWilcurt:
@svenjunke That is my feeling as well. Just make one normalize.sass file and it will automatically create the normalize.css and normalize.min.css for the non-sass users.

fleeting:
I still don't understand the need in this being sass. I get normalize with bower, then have gulp copy it as normalize.scss, which is then compiled with my sass files. You could also just have your build process combine normalize.css with your css instead.

I guess I'm just missing the need for this besides having it in sass, just to have it in sass.

florianbouvot:
@fleeting +1
We don't need Sass version, Sass is CSS compatible so you can create your own build process.

stevendchen:
from necolas: 
> "Thanks for taking the time to do this. There have been a couple of pull requests related to Sass versions of the file, but I'd rather not have to maintain 2 near-identical files. I'm happy to either let people Sass-ify the file themselves or have others maintain a Sass version."

if you'd really like there's:
https://github.com/esses/normalize.sass

and 
https://github.com/JohnAlbin/normalize-scss
https://github.com/guerrero/normalize.scss

but as @fleeting said, if you're using gulp or the like you can just stick it in to your build as css

necolas:
I still don't see why this would be a good thing to do :)

Issue #290, opened 2014-02-03
normalizer.scss : The Normalize ( CSS ) Library for Sass


stef-rausch:
Hi @necolas,

I just have released [normalizer.scss](https://github.com/StefanoRausch/normalizer.scss) v3.0.0, which might let you rethink your standpoint about not using Sass as an additional way of distribution. Nothing fancy, sticks to your original file, but is highly flexible — hint : it only uses placeholder selectors and no mixins etc.

Thanks a lot for `normalize.css`!

PR #557, opened 2016-03-23
Add sass file

ppsirius:
On github we can find many repo with normalize in sass, why don't make this file on official repository?
+ file ready to import
+ output css a half smaller than file with comments

### DIFF
--- a/normalize.scss
+++ b/normalize.scss
@@ -1,0 +1,263 @@
+/*! normalize.css v4.0.0 | MIT License | github.com/necolas/normalize.css */
+
+// 1. Change the default font family in all browsers (opinionated).
+// 2. Prevent adjustments of font size after orientation changes in IE and iOS.
+html {
+  font-family: sans-serif; // 1
+  -ms-text-size-adjust: 100%; // 2
+  -webkit-text-size-adjust: 100%; // 2
+}
+
+// Remove the margin in all browsers (opinionated).
+body {
+  margin: 0;
+}
+
+// HTML5 display definitions
+// ==========================================================================
+
+// Add the correct display in IE 9-.
+// 1. Add the correct display in Edge, IE, and Firefox.
+// 2. Add the correct display in IE.
+article, aside, details, figcaption, figure, footer, header, main, menu, nav, section, summary {
+  display: block;
+}
+
+// Add the correct display in IE 9-.
+audio, canvas, progress, video {
+  display: inline-block;
+}
+
+// Add the correct display in iOS 4-7.
+audio:not([controls]) {
+  display: none;
+  height: 0;
+}
+
+// Add the correct vertical alignment in Chrome, Firefox, and Opera.
+progress {
+  vertical-align: baseline;
+}
+
+// Add the correct display in IE 10-.
+// 1. Add the correct display in IE.
+template, [hidden] {
+  display: none;
+}
+
+// Links
+// ==========================================================================
+
+// Remove the gray background on active links in IE 10.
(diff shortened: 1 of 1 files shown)


battaglr:
@ppsirius, hi! Thank you for taking the time to do this! Preprocessors or transpilers such as Sass have been the topic of many issues, such as #450, #446, #290, and [the list goes on and on](https://github.com/necolas/normalize.css/issues?utf8=%E2%9C%93&q=sass).

Before releasing v4 we considered this, since it has been asked many times, but we really couldn't find what value could it bring.

In all preprocessors/transpilers that I know of, you can import a vanilla CSS; and while I agree "silent" comments are great, that doesn't justify maintaining two different files. If we decided to merge this, what about Less or Stylus, or any flavour of PostCSS, which would be our argument to not accept them —and end up maintaining lots of different files with mostly different extensiones?

ppsirius:
I agree with you, You should add other preprocessors. This files give us possibility to clean fast comments in output. Reducing the file size from 7kb to 3kb i think is enough reason. Otherwise on github you will see a bunch of duplicated repo.

TrySound:
@ppsirius If you care about style size use css minifiers. If you use preprocessors I think it's not so big deal to add something in pipe.

TrySound:
@ppsirius Your way is opposite to DRY.

greypants:
I think for a while trying to `@import` css files would default to the native css `@import` and link to an external file rather than pull in the file directly like it does with scss/sass.  

This was [fixed in `node-sass`](https://github.com/sass/libsass/pull/754) and now works fine if you leave off the `.css` extension.

I added this to my sass options:

```js
includePaths: [
  "./node_modules/normalize.css"
]
```

And can `@import` the `normalize.css` file just fine in my main `.sass` file.

```sass
@import normalize
```

**HOWEVER**, @battaglr, [this is still an issue in Ruby Sass](https://github.com/sass/sass/issues/556), and duplicating `normalize.css` as `normalize.scss` in this repo, would fix the problem for a lot of people. This project (normalize) was the use case mentioned in the issue.

It's a super easy add (`cp normalize.css normalize.scss`), and like @ppsirius said, would remove the need for forks like https://github.com/JohnAlbin/normalize-scss

battaglr:
@ppsirius, please, keep in mind that *development code* is not the place for optimizations such as this one —if you're concerned with size, as @TrySound suggested, in your *production code* you may consider using a minifier. Please, [don't forget that the detailed comments are an important aspect of this project](https://github.com/necolas/normalize.css#what-does-it-do).

@greypants, hi! Thanks for correcting me on this; I didn't know Ruby Sass had this behaviour. I'm not in charge of making any decisions, but regardless of how easy or difficult including this here could be, usually the best/right thing to do is to put the energy on solving the *problems or limitations* of a given project in said project, and not trying *patch* other projects because said *problems or limitations*. FWIW @chriseppstein himself wrote a `gem` to allow *fix* that: [sass-css-importer](https://github.com/chriseppstein/sass-css-importer).

jonathantneal:
It’s a noble idea, but if or when we introduce a Sass, Scss, Stylus, or Less version, or any other version, it will be generated from a true CSS source. Perhaps PostCSS can do this. May all of these extensions be generated in PostCSS?

For now, we must not repeat ourselves. I allowed for this in [sanitize.css](https://github.com/10up/sanitize.css/tree/master/lib) and now upgrades are a beast. So, I have experience going the other way, and I can say with confidence that it has hurt the project.

PR #571, opened 2016-04-07
Correct the text style of placeholders in Chrome, Edge, and Safari

### LINKED ISSUE #550: Placeholders are not consistent in all browsers
IE, Edge and Firefox have different placeholders colors than other browsers
Should add:
:-ms-input-placeholder {
	color: #a9a9a9;
}
::-ms-input-placeholder {
	color: #a9a9a9;
}
:-moz-placeholder,
::-moz-placeholder {
	color: #606060;	/* -> #a9a9a9 on display */
}

jonathantneal:
Normalizes placeholder styles in Edge, Chrome, and Safari with Firefox.

Resolves #550

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -393,3 +393,12 @@ textarea {
 [type="search"]::-webkit-search-decoration {
   -webkit-appearance: none;
 }
+
+/**
+ * Correct the text style of placeholders in Chrome, Edge, and Safari.
+ */
+
+::-webkit-input-placeholder {
+  color: inherit;
+  opacity: 0.54;
+}


Yapaslfeu:
[https://github.com/necolas/normalize.css/issues/550#issuecomment-208262491]

> I think the IE problem is a big issue for the moment, as it's still used a lot!
Despite all the discussions, i still believe that we have to force the placeholder color to a light grey for the moment, and fix Firefox - the only browser which do it differently - to opacity:1!

Firefox is not necessary right: light silver is like a system color for placeholders, and placeholders with a lighter color than the input text is not so logic in my point of view.

Above all, it's far more easy for one to change all the placeholders colors if needed (i.e. bad readability, which is not so common and immediately visible, even at design phase), than obliging everybody to set the placeholder color with a reba when changing the fields text color (very common modification), this only for IE 11!!

Yapaslfeu:
At least, if you are not convinced by my comment, you need to set the placeholder color for IE, as it doesn't use the text color by default:
`:-ms-input-placeholder {`
`	color: inherit;`
`}`

PR #573, opened 2016-04-08
Remove gaps in links underline in iOS 8+ and Safari 8+

hudochenkov:
Remove gaps in links underline in iOS 8+ and Safari 8+

This gaps presented in Safari and iOS only. They were added in Safari 8 and iOS 8 but CrossBrowserTesting show them even in Safari 6.

[CodePen without fix](http://codepen.io/hudochenkov/pen/RaxqNW) and [results](https://app.crossbrowsertesting.com/public/i1769107ade2a10c/screenshots/z894d678ffb8d28e7e2b).

screen shot 2016-04-08 at 21 33 08

[CodePen with fix](http://codepen.io/hudochenkov/pen/groQbM) and [results](https://app.crossbrowsertesting.com/public/i1769107ade2a10c/screenshots/z5cba15df917b55d4a30).

screen shot 2016-04-08 at 21 33 10

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -85,11 +85,13 @@ template, /* 1 */
    ========================================================================== */
 
 /**
- * Remove the gray background on active links in IE 10.
+ * 1. Remove the gray background on active links in IE 10.
+ * 2. Remove gaps in links underline in iOS 8+ and Safari 8+.
  */
 
 a {
-  background-color: transparent;
+  background-color: transparent; /* 1 */
+  -webkit-text-decoration-skip: objects; /* 2 */
 }
 
 /**


jonathantneal:
I think this change is intentional.

https://css-tricks.com/almanac/properties/t/text-decoration-skip/
https://medium.com/designing-medium/crafting-link-underlines-on-medium-7c03a9274f9#.64aes54bm

Which is happy and sad. Happy because it improves readability of text. Sad because this is one of the most perfect PRs I’ve seen.

hudochenkov:
Thank you for the kind words :)

I think removing these gaps is making link underlines consistent between browsers which is the main purpose of normalize.

jonathantneal:
You are correct that normalize.css is about making things consistent, but not necessarily to the lowest common denominator. For instance, normalize.css v1 would not have added `* { box-shadow: none !important; }` to prevent browsers from looking different from IE8. The exception to this rule was when the property impacted the box model. For instance, Firefox used to have a line height on inputs that could not be overridden, so normalize.css prevented all browsers from changing this.

However, I need to look into this further, because [the spec](https://drafts.csswg.org/css-text-decor-3/#text-decoration-skip-property) indicates that the default is `objects`, while Safari is defaulting to `ink`.

thierryk:
I think it would make more sense to "normalize" this using the initial value (`objects`) rather than `none`.

jonathantneal:
@hudochenkov, if you would be open to modifying your commit to change `none` to `objects`, I would take in this PR for 4.1.0.

hudochenkov:
@jonathantneal I updated commit. [Codepen with example](http://codepen.io/hudochenkov/pen/QNQdxm).

jonathantneal:
@hudochenkov, @thierryk, thank you both!

PR #574, opened 2016-04-10
Normalize file select buttons

### LINKED ISSUE #570: No uniform aspect of "file select" buttons after normalization
In the same way you are setting -webkit-appearance:button in line 299, i think you should do the same for the "select file" button present in file fields, in order to keep a uniform aspect between all fields in Safari:

[type='file']::-webkit-file-upload-button {
-webkit-appearance: button;
}

jonathantneal:
> In the same way you are setting -webkit-appearance:button in line 299, i think you should do the same for the "select file" button present in file fields, in order to keep a uniform aspect between all fields in Safari:

Resolves #570

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -291,7 +291,7 @@ select { /* 1 */
 /**
  * 1. Prevent a WebKit bug where (2) destroys native `audio` and `video`
  *    controls in Android 4.
- * 2. Correct the inability to style clickable types in iOS.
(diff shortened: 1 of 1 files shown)


slavanga:
The webkit selector will make non-webkit browsers ignore the whole rule.

Yapaslfeu:
And the font:inherit is not useful as all browsers i tried already apply the inherited font in the file selector button

jonathantneal:
Thanks, @slavanga!

PR #575, opened 2016-04-10
Correct the outline of search inputs in Chrome and Safari

### LINKED ISSUE #547: Search input focus outline not correct in Safari
After modifying search input with -webkit-appearance: textfield, the focus outline is remaining with outline-offset=0, but should be -2px like other textfield inputs:
[type='search']:focus {
	outline-offset: -2px;
}

jonathantneal:
After modifying search input with -webkit-appearance: textfield, the
focus outline is remaining with `outline-offset: 0`, but should be
`-2px` like other textfield inputs.

Resolves #547

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -381,11 +381,13 @@ textarea {
 }
 
 /**
- * Correct the odd appearance of search inputs in Chrome and Safari.
+ * 1. Correct the odd appearance in Chrome and Safari.
+ * 2. Correct the outline style in Safari.
  */
 
 [type="search"] {
-  -webkit-appearance: textfield;
+  -webkit-appearance: textfield; /* 1 */
+  outline-offset: -2px; /* 2 */
 }
 
 /**


Yapaslfeu:
Sorry, i have the (bad?) habit to use single quotes for CSS strings…

normalize.css
```diff
@@ -387,6 +387,14 @@ textarea {
 }
 
 /**
+ * Correct the outline of search inputs in Chrome and Safari.
+ */
+
+[type='search']:focus {
```
FagnerMartinsBrack: All other lines uses double quote, shouldn't this one be consistent?

jonathantneal:
This is why we have integrated testing. :wink:

thierryk:
It says this is for Chrome and Safari, but in Chrome, it seems the UA style sheet contains this rule already:

```
input:focus, 
textarea:focus, 
keygen:focus, 
select:focus {
    outline-offset: -2px;
}
```

So in my opinion the above should take care of the issue in Chrome. Isn't the case?

thierryk:
There is definitely something going on in [Safari with `[type="search"]`](http://codepen.io/thierry/pen/XdZzGw) (test on codepen)

thierryk:
Anyway... this looks good to me besides the fact that the comment should read:

```
Correct the outline of search inputs in Safari.
```

instead of:

```
Correct the outline of search inputs in Chrome and Safari.
```

If indeed the issue is _limited_ to Safari only

jonathantneal:
I was able to take a real quick look, confirm what @thierryk said, and update the commit.

mac_osx_10 11_safari_9_1024x768_chromeless

thierryk:
@jonathantneal that should do it but please note that it is only because Safari seems to be broken there too as the UA style sheet contains the following rule:

```
input[type="search"] {
    -webkit-appearance: searchfield;
    box-sizing: border-box;
}
```

The above clocks at `0.0.1.1` versus `0.0.1.0` for the selector used in this commit. So technically it should not overwrite that style from the UA style sheet (even though it does).

jonathantneal:
I think that’s because UA styles have zero weight, regardless of the selector.

thierryk:
> I think that’s because UA styles have zero weight, regardless of the selector.

I'm sure you're right (after all these years I've never noticed that, duh!).

PR #563, opened 2016-03-26
Remove opinionated cursor styles on buttons

jonathantneal:
Resolves #544 and #371

PR #585, opened 2016-04-15
Add role="button" to fix common iOS Safari bug

smartmike:
Just came across this little bug on a project, and figured it was something that could be helpful to others.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -417,3 +417,11 @@ textarea {
   -webkit-appearance: button; /* 1 */
   font: inherit; /* 2 */
 }
+
+/**
+ * Allow elements with role=button to be pressed in Mobile Safari
+ * See: https://developer.mozilla.org/en-US/docs/Web/Events/click#Safari_Mobile
+ */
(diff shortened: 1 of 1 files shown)


jonathantneal:
Hi, @smartmike. normalize.css once made that same decision, but after a lot of wisdom from the larger community, we pulled those styles. You can read about it at https://github.com/necolas/normalize.css/issues/371#issuecomment-60072171 and see the PR that made the change to remove these styles at https://github.com/necolas/normalize.css/pull/563

While changing the pointer may or may not be the best decision in your project, they are definitely opinions as opposed to normalizations, and ultimately they may be harmful in some situations.

In my own experience, and this is my own opinion, I find that people expect pointer: cursor during QA, but in reality most folks mostly rely on the button being something obviously pressable or clickable (depending on their input) before they ever put their finger or mouse over it, and then they rely on the hover or active state to validate that interaction.

smartmike:
I see your point. Thanks for the input, and taking the time to answer!

PR #588, opened 2016-04-17
Re-ordered for specificity

matt3224:
https://jonassebastianohlsson.com/specificity-graph/

Here's an example of a specificity based normalize. I'm submitting it to open a discussion more than anything else.

Although this may feel somewhat extreme, I personally would much prefer a few duplicate properties and a nice specificity trend over fewer properties and a spikey graph.

I may try my hand at a specificity grouped reset, as opposed to grouping by element type.

Is specificity worthwhile in a reset, or is the concern more with file size?

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -19,9 +19,6 @@ body {
   margin: 0;
 }
 
-/* HTML5 display definitions
-   ========================================================================== */
-
 /**
  * Add the correct display in IE 9-.
  * 1. Add the correct display in Edge, IE, and Firefox.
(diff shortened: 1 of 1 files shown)


thierryk:
For what it is worth, I think "specificity graphs" are a complete joke. Here it is, I said it :)
I don't think redundancy (which matters to browsers) should be introduced to produce a "nice graph" (which does not matter to browsers one bit).

matt3224:
Fair enough @thierryk, I guess you should be aware of specificity, bare it in mind, but don't bend to its will. Its good to get a different perspective on these things.

thierryk:
Ordering only matters for styles of *same weight* as the cascade will favor the ones coming last. For other styles it does not matter what comes first. Actually, you could have a graph going in the **opposite** direction (down instead of up) and still have everything working as expected.

PR #594, opened 2016-05-31
remove fieldset margin

lunix01:


battaglr:
@lunix01, hi! Would you mind to elaborate on the reasons behind this change? :smiley:

Issue #202, opened 2013-06-05
Use box-sizing:border-box for all elements


rpk103:
Although this would be a fairly major enhancement from a testing perspective, I think it would be forward-looking to use box-sizing:border-box for all elements (*) as the default and override it only for those places where there are browser inconsistencies.  Coverage for this feature is really good across modern browsers and developers should be using it for all new HTML5 development.  Not incorporating it in normalize.css makes it difficult to use for all elements in the app stylesheet and still get the benefits of using normalize.css.   Seeing this take off in HTMLBoilerplate (through their use of normalize.css) would really spread the benefits of this best practice.

See the Paul Irish link below for some context on the topic.

http://www.paulirish.com/2012/box-sizing-border-box-ftw/

PR #598, opened 2016-06-25
border-box added

kvnol:
Correct border with border-box

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -1,6 +1,18 @@
 /*! normalize.css v4.1.1 | MIT License | github.com/necolas/normalize.css */
 
 /**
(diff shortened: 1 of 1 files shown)


battaglr:
Please, see #202 and #409.

jonathantneal:
To summarize. These kinds of changes are outside the scope of normalize.css. No specs have this. No browsers do this. It's easy enough to add on your own. Push opinionated defaults to [sanitize.css](https://github.com/10up/sanitize.css).

thierryk:
> Correct border with border-box

As a side note, `border-box` may be a *preferred* way for some authors but it is certainly not the "correct" way.

Issue #593, opened 2016-05-04
h1 line height different in Safari and Firefox (last version)


jeremenichelli:
I've noticed a difference in the line height of  `<h1>` tags between these two and I've created this issue to know if normalize tries to mitigate this kind of differences too.

In case maintainers feel it should I can take care of creating the pull request. Thank you.

Karklis:
if you zoon out on Firefox to 80% show the same size as in Safari or Chrome?

jeremenichelli:
@Karklis nope, in Safari and Chrome the `line-height` is `50px` and in Firefox it's `39.4px' and zooming out barely increases it to `39.9px`

Karklis:
Could you provide code and all steps to reproduce this behavior? Are you using just default normalize css without any other css code, or mix something?

jeremenichelli:
Create a new page with this content:

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Line height diff in h1 tags</title>

    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/normalize/4.1.1/normalize.min.css">
</head>
<body>
    <h1>line height diff in h1 tags</h1>

    <p>line-height: <span id="lh"></span></p>
    <script>
        document.querySelector('#lh').textContent = window.getComputedStyle(document.querySelector('h1'))[ 'lineHeight' ];
    </script>
</body>
</html>
```

You will be able to see the computed line height on page or open the web inspector in Safari and Firefox and check the resulting computed styles for `line-height` property.

### Results

**Firefox:** `33px`
**Safari:** `37px`

Considering that in both the `font-size` is `32px`, conclusions point that Firefox is assigning only `1.03em` of `line-height` while Safari and Chrome `1.15625em`.

jeremenichelli:
Hi @Karklis, were you able to test this? Do you prefer another presentation for the test case?

jonathantneal:
I’ve put your test on CodePen. I see a difference between Chrome and Firefox, between Firefox Mac and Firefox PC, and possibly by a fraction of a pixel in Edge/IE.

http://codepen.io/jonneal/details/RRNpLB/

jeremenichelli:
So, what do we take as a standard to make them all match? I'll create the
PR :)

On Tuesday, 31 May 2016, Jonathan Neal <notifications@github.com> wrote:

> I’ve put your test on CodePen. I see a difference between Chrome and
[...]

PR #601, opened 2016-06-25
Simplify headings normalization comment

battaglr:
The results for computed `line-height` in #593 were virtually different in all tested browsers, and since [such browsers are the ones we support](https://github.com/necolas/normalize.css#browser-support), I don't see the need to have a *complex* comment naming all of them. Also, having the "headings" reference is redundant, since by looking at the selectors is clear what we are targeting.

I know this is kinda silly, but it's important to keep comments short and to the point. 😉

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -155,8 +155,7 @@ h1 {
 }
 
 /**
- * Correct the line-height for all headings in Chrome mobile, Firefox,
- * iOS Safari, Microsoft Edge and IE.
+ * Correct the line height for headings in all browsers.
  */
 
 h1,
--- a/normalize.css
+++ b/normalize.css
@@ -155,7 +155,7 @@ h1 {
 }
 
 /**
- * Correct the line height for headings in all browsers.
+ * Correct the line height in all browsers.
  */
 
 h1,


thierryk:
@battaglr Are the results different for headings *only* or is that discrepancy due to *different* inherited values across browsers (i.e. different computed values for `line-height:normal`).

jonathantneal:
@battaglr, the issue is with `line-height: normal` in all browsers, not just headings. Normalization should probably occur on the root `html` element. Here are some test results https://app.crossbrowsertesting.com/public/i76b092cd2b52b86/screenshots/z258aeb78893f4a768e5

And here is the test: http://codepen.io/jonneal/pen/RRRJRp

battaglr:
@thierryk, I think that @jonathantneal already answered your question. :grinning:

1) In this PR I'm only suggesting to simplify an existing comment for code that only targets headings. *Since the diff doesn't show it, I'll add the full rule for context:*

```css
/**
 * Correct the line height in all browsers.
 */

h1,
h2,
h3,
h4,
h5,
h6 {
  line-height: 1.15;
}
```

2) According to the test results that @jonathantneal provided, the `line-height` needs to be normalized for all elements. If adding `line-height: 1.15;` to the `html` element solves that, I'll happily modify this PR —or open a new one— removing the `h1` to `h6` rule and adding the one that targets the `html`. What do you guys think?

thierryk:
@battaglr Yeah, @jonathantneal and I have had that discussion elsewhere already ;)
I think best is to normalize `line-height` on `html`. Regarding the value, we'd need to use whatever `normal` compute to in most browsers. If it's `1.15` then be it.

battaglr:
@thierryk, great, we agree on that. Let's wait for @jonathantneal opinion.

jonathantneal:
@battaglr, yes, this should only target `html`.

battaglr:
@jonathantneal, awesome! Thanks for merging with the change already in! It's great that this change was originated by a silly change in a comment (so thanks @thierryk for questioning what at first I considered "unrelated" to the PR).

PR #603, opened 2016-06-30
Remove `placeholder` normalization

### LINKED ISSUE #580: Correct the text style of placeholders in IE
Not sure you get my comments on #571:

> At least, if you are not convinced by my comment, you need to set the placeholder color for IE, as it doesn't use the text color by default:
`:-ms-input-placeholder {`
`	color: inherit;`
`}`

battaglr:
We can not safely normalize the `placeholder` without introducing a specific color —which would be very opinionated—, or without modifying the `opacity` —which would prevent the placeholder from being shown in Edge under certain conditions [due to a bug](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/3901363/).

*Context*: this was originally discussed on #550 and then added on #571. The decision to remove it was made on #580.

*Notes*: (a) I didn't modify the `test.html` file since I couldn't find a test for this. (b) I didn't add a note to the README as @jonathantneal suggested since as we no longer include this, it would be weird to add it to a section with the "additional detail and explanation of the esoteric parts of normalize.css." tagline —please, feel free to add it if you deem it necessary.

---

This closes #580.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -403,15 +403,6 @@ textarea {
 }
 
 /**
- * Correct the text style of placeholders in Chrome, Edge, and Safari.
- */
-
-::-webkit-input-placeholder {
-  color: inherit;
-  opacity: 0.54;
-}
-
-/**
  * 1. Correct the inability to style clickable types in iOS and Safari.
  * 2. Change font properties to `inherit` in Safari.
  */


slavanga:
Looks good. Can we get a proper release after the merge?

battaglr:
@slavanga, I think that it would be the best thing to do —at least a *minor* release—, to prevent people from being caught with a weir Egde bug.

Yapaslfeu:
You did a very good resume of the situation, but majority of browsers use the same grey color for placeholders, so specifying a placeholder color is not so arbitrary…

Safari & Chrome use #A9A9A9
Edge & IE use #6D6D6D
Firefox use #767676 (black with .54 opacity)

alrra:
Somehow similar discussion: https://github.com/h5bp/html5-boilerplate/pull/1403.

jonathantneal:
@visicode, I’d like to avoid setting those explicit values here because of how it alters the expectation for people leveraging opacity. To provide the best normalized experience across all browsers, [some kind of PostCSS plugin](https://github.com/necolas/normalize.css/issues/550#issuecomment-206887823) would be needed. I would also recommend something like https://github.com/visicode/formalize.css and I realize now this is your own project. This would be a great addition to your opinionated project. If focused strictly on forms, I think it could be a great companion to both normalize.css and sanitize.css.

battaglr:
@alrra, great data! Thanks for linking the issue, would be great for future references.

jonathantneal:
Thanks!

battaglr:
@jonathantneal, great! The note in the README is a nice addition, thanks!

*Off-topic:* we have so much knowledge accumulated on issues and PRs that creating another document (even another project) listing all that it's starting to sound like a good idea.

Issue #604, opened 2016-07-02
Set <summary> elements to display:list-item


Mottie:
The disclosure triangle of `<summary>` blocks does not show in Firefox when it is set to `display:block`; According to the bugzilla report (linked below), this is the expected behavior. Therefore it's default setting should be set as `display:list-item` (CC @silverwind).

References:
* https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d
* https://bugzilla.mozilla.org/show_bug.cgi?id=1283989
* https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements

silverwind:
I think this is more of an GitHub bug than anything, they explicitely set `summary {display:block}` which causes the incorrect rendering on Firefox.

Mottie:
If you look at [GitHub's current css source](https://assets-cdn.github.com/assets/frameworks-fef7fdfe41dd435c6d82f93b486dc62bc095ebd12cd24fa2a817c8bb45dc073e.css), they are using:

```css
/*! normalize.css v4.1.1 | MIT License | github.com/necolas/normalize.css */
```

silverwind:
Ah, didn't realize that `display: block` is happening in normalize.css, so it's a valid issue indeed!

jonathantneal:
Thanks for letting me know, and also for the link in the spec. I’ll take care of this as soon as I can validate this in a test. Looks like the new display property will work just fine per https://developer.mozilla.org/en-US/docs/Web/CSS/display#compat-desktop (but again, testing to be sure)

jonathantneal:
**Implementation status**:
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1283989
Chrome: https://bugs.chromium.org/p/chromium/issues/detail?id=590014

**Browser “Last 2” test results**: https://app.crossbrowsertesting.com/public/i76b092cd2b52b86/screenshots/z344607e8fe6c4982ab8

**Pull Request**:
https://github.com/necolas/normalize.css/pull/606

Mottie:
Odd... the internal Firefox stylesheet (resource://gre-resources/html.css) sets summary with a `display:block` initially. But if it is wrapped inside a `<details>` element, this definition applies:

```css
details > summary::-moz-list-bullet {
  list-style-type: disclosure-closed;
  /* Prevent elements in summary being selected when clicking on the triangle. */
  -moz-user-select: none;
}

[...]

PR #606, opened 2016-07-06
Use HTML Living Standard sectioning

jonathantneal:
The 1st commit:
1. Adds any normalized sections not already present from https://html.spec.whatwg.org/multipage/
2. Moves unsorted rules into their respective sections.

(which makes no changes to rendering, but is basically a setup for)

The 2nd commit:
1. Corrects the display for `<summary>` in all browsers, resolving #604.

@battaglr, would you have a moment to review?

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -6,12 +6,18 @@
  * 3. Prevent adjustments of font size after orientation changes in IE and iOS.
  */
 
+/* Document
+   ========================================================================== */
+
 html {
   font-family: sans-serif; /* 1 */
   line-height: 1.15; /* 2 */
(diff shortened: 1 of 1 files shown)


battaglr:
@jonathantneal

1) Using the new categories to define the code organization it's a nice improvement. The only drawback is that some elements may belong to more than one category, making us have to choose just one and maybe adding some confusion. But I think it's not a problem, just wanted to mention it just in case.

2) The code looks good, but there are some new interesting comments in #604 that are worth checking.

Toddses:
#695

PR #607, opened 2016-07-07
More explicitly define font resets on form controls

### LINKED ISSUE #591: input font: inherit also pulls in line height
I have a search form where the parent element sets a line height to vertically centre the text in the form. The following rule brings that line height to the text inside the input in the form:

button,
input,
optgroup,
select,
textarea {
  font: inherit; /* 1 */
}

Perhaps this was overlooked? I don't think it's a great 'normalisation' as I don't think anyone would expect this particular aspect of the font style to come through. Perhaps we could be more specific and inherit just the font-face?

jonathantneal:
Resolves #591 by explicitly setting `font-family`, `font-size`, and `line-height` rather than using `inherit`.

^ @battaglr, also see the `line-height` tests in the related issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 * Move unsorted rules into their respective sections.
 * Update the `summary` style in all browsers.
 * Remove `::placeholder` styles due to a bug in Edge.
+* More explicitly define font resets on form controls
+* Remove the `optgroup` normalization needed by the previous font reset
 
 ### 4.2.0 (June 30, 2016)
 
--- a/normalize.css
+++ b/normalize.css
@@ -244,7 +244,7 @@ svg:not(:root) {
    ========================================================================== */
 
 /**
- * 1. Change font properties to `inherit` in all browsers (opinionated).
+ * 1. Change the font styles in all browsers (opinionated).
  * 2. Remove the margin in Firefox and Safari.
  */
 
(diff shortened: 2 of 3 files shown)


battaglr:
@jonathantneal, the code looks great (and thanks for the tests)! I think there's a couple of things that we are missing:

- Remove the [`optgroup` normalization](https://github.com/necolas/normalize.css/blob/master/normalize.css#L260-L266), as this change makes it unnecessary, right?
- We probably have to update [the font test](https://github.com/necolas/normalize.css/blob/master/test.html#L261-L273) to match the new changes.
- What are we going to do with [`::-webkit-file-upload-button`](https://github.com/necolas/normalize.css/blob/master/normalize.css#L414), should we remove `font` in favour of these new properties?

avastamin:
* {
    -webkit-box-sizing: border-box;
    -moz-box-sizing: border-box;
    box-sizing: border-box;
}  

You should add this as well.

jonathantneal:
@avastamin, we all love `box-sizing: border-box`, but that is not a browser default anywhere, and it has nothing to do with this pull request. For something like normalize.css with developer preferences built in try [sanitize.css](https://github.com/jonathantneal/sanitize.css).

@battaglr, I’ll update the `optgroup` normalization. Thank you for finding that. I will leave `::-webkit-file-upload-button` until we can discuss it in an issue first.

battaglr:
@jonathantneal, great! I agree that we can discuss things like `::-webkit-file-upload-button` in another issue to avoid delaying this PR. 👍

PR #615, opened 2016-08-08
Feature/semver strategy

### LINKED ISSUE #612: Global line-height rationale, semver strategy revision
What's the rationale behind `html { line-height: 1.15 }`? I thought it was supposed to be [just the headings](https://github.com/necolas/normalize.css/issues/593)? And wouldn't such a drastic change warrant a major version bump?

jonathantneal:
This change updates our semver strategy to reflect and resolve #612. If this is merged, it would be good to then push a new **major** release.


battaglr:
@jonathantneal, LGTM! Let's see how this works. 😀

PR #626, opened 2016-09-13
Update normalize.css

mattbrundage:
Clarification that this rule applies to IE's mobile variant. Interestingly, we'll be able to remove "-ms-text-size-adjust" soon, as Windows Phone 8.1+ supports "-webkit-text-size-adjust" as an alias.

Documentation: https://msdn.microsoft.com/en-us/library/dn793579(v=vs.85).aspx

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -3,7 +3,7 @@
 /**
  * 1. Change the default font family in all browsers (opinionated).
  * 2. Correct the line height in all browsers.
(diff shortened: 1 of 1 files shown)


jonathantneal:
Thanks! Merged via https://github.com/necolas/normalize.css/commit/f9b42c6a3bd89ec6d4e1506051cbc63128802dcb

PR #628, opened 2016-09-27
Create normalize.scss

aziev:
It's just a copy of normalize.css.
After downloading `normalize` via `npm` it creates only `normalize.css`, so developer needs to change the file extension each time because `SCSS` supports only `.scss` files import.


kevva:
https://github.com/necolas/normalize.css/search?q=scss&type=Issues&utf8=%E2%9C%93

PR #636, opened 2016-10-08
Changed ButtonColor to inherit.

colshacol:
Not quite sure if this line was meant just to cancel out any color added, or if it used to be a variable in some preprocessor.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -313,7 +313,7 @@ button:-moz-focusring,
 [type="button"]:-moz-focusring,
 [type="reset"]:-moz-focusring,
 [type="submit"]:-moz-focusring {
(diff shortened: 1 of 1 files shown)


mattbrundage:
ButtonText is a [system color](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#System_Colors) and its use here in intentional.

adamralph:
https://jigsaw.w3.org/css-validator/ tells me

> The value 'ButtonText' is deprecated

jonathantneal:
What @mattbrundage said :)

adamralph:
From the link posted by @mattbrundage:

>  Not all system colors are supported on all systems. 👎 for use on public web pages.

PR #639, opened 2016-10-17
Place the document heading above the html element's comment.

JohnAlbin:
The section heading comment always appears above the individual comments for rulesets in that section.

Except for the new "Document" heading.

This PR fixes that.


battaglr:
@JohnAlbin, great! 😀

PR #643, opened 2016-11-17
add pointer-events: none for svg:not(:root)

jzilg:
In Firefox inline SVGs causes JavaScripts Bugs when hovering with the cursor. Setting the pointer-events to none is the workaround for that.

https://css-tricks.com/links-inline-svg-staying-target-events/

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -234,11 +234,13 @@ img {
 }
 
 /**
- * Hide the overflow in IE.
+ * 1. Hide the overflow in IE.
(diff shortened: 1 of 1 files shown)


thierryk:
I'm not sure this "fix" belongs to *normalize*, but I could be wrong...

Llbe:
This is not a bug in Firefox. And nothing on css-tricks.com says it is.

IMO, this should not be in _normalize_.

jonathantneal:
Not a normalization.

PR #644, opened 2016-11-24
Change the font style also for keygen in Chrome

jKratzik:
Styling the keygen tag itself has no effect

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -253,7 +253,8 @@ button,
 input,
 optgroup,
 select,
-textarea {
+textarea,
+keygen::-webkit-keygen-select {
   font-family: sans-serif; /* 1 */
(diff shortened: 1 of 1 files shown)


thierryk:
Not sure if we should care much about this as I believe `<keygen>` is being deprecated.

jKratzik:
Yes right.

There is a discussion going on but it seems that the tag will be removed: https://developer.mozilla.org/de/docs/Web/HTML/Element/keygen
And Chrome is going to drop it: https://www.chromestatus.com/feature/5716060992962560

jonathantneal:
Because the tag is deprecated and the selector is heavier than a plain tag, I would like to avoid adding this decision fatigue to the project. Thank you, though, for a wonderfully clear pull request.

PR #661, opened 2017-03-23
Update normalize.css

mattbrundage:
Comment upkeep, per https://bugs.chromium.org/p/chromium/issues/detail?id=376160#c11

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -118,8 +118,8 @@ a:hover {
 }
 
 /**
- * 1. Remove the bottom border in Firefox 39-.
- * 2. Add the correct text decoration in Chrome, Edge, IE, Opera, and Safari.
+ * 1. Remove the bottom border in Chrome 57- and Firefox 39-.
+ * 2. Add the correct text decoration in Edge, IE, Opera, and Safari.
  */
 
 abbr[title] {


jonathantneal:
This was super helpful and really polished the v6 release. Thanks, @mattbrundage !

nekolab:
Hi, I'm not sure about change in this commit, since it seems no bottom border in chrome 57- in my mind.

mattbrundage:
@nekolab, if [this commit](https://chromium.googlesource.com/chromium/src/+/fb579cdb6aa334f44e57072a3c1fd803aa2cb082%5E%21/third_party/WebKit/Source/core/css/html.css) is any indication, then you're right -- there was no prior `border-bottom` rule in older versions of Chrome.

Support for the new text-decoration properties [landed in Chrome 57](https://bugs.chromium.org/p/chromium/issues/detail?id=665898), while default `abbr` styling [landed in Chrome 58](https://bugs.chromium.org/p/chromium/issues/detail?id=376160). Perhaps we should just omit mention of Chrome from these two `abbr` comments, given the closeness of these two feature landings..

nekolab:
@mattbrundage Yeah, what about revert change in this commit and change comment 2 to
`2. Add the correct text decoration in Chrome 58-, Edge, IE, Opera, and Safari.`?

PR #669, opened 2017-04-25
Add acronym tag

Habetdin:
Acronym is almost the same that abbr is.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -102,7 +102,7 @@ a {
  * 2. Add the correct text decoration in Chrome, Edge, IE, Opera, and Safari.
  */
 
-abbr[title] {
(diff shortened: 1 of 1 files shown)


thierryk:
According to [https://www.w3.org/wiki/HTML/Elements/acronym]():

> The `<acronym>` element is a non-standard element.
> HTML5 classifies it as a non-conforming feature.

[11.2 Non-conforming features](https://www.w3.org/TR/html5/obsolete.html#non-conforming-features):

> Elements in the following list are entirely obsolete, and must not be used by authors:
> [...]
> `acronym`
> Use `abbr` instead.

Habetdin:
Hmm... Didn't know it's obsolete atm :disappointed:

PR #671, opened 2017-04-28
Fieldset padding in firefox

rellect:
v6 has removed the opinionated styling for the `fieldset` element.
One of the removed rules was the padding, and I noticed that this introduced a little inconsistency between firefox and other browsers.
In my test, I saw consistent padding in chrome and IE. But a slightly different padding in Firefox.

Firefox
image

Chrome
image

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -293,6 +293,14 @@ button:-moz-focusring,
 }
 
 /**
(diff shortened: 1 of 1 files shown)


jonathantneal:
Thanks @rellect !

Issue #664, opened 2017-03-31
On the "Opinionated rules"


ahmadalfy:
I understand that Normalize isn't a classical CSS Reset but removing those that you call _opinionated rules_ is harsh. It's 2017 everybody is setting their box sizing to border-box. Leaving the body margin? How many websites don't have this set to zero?

For me Normalize is considered a reset for the non useful stuff like the body margin and a way to make the different browsers behave consistently. I don't really like the new update and I am sure that there are a lot who disagree with it.

Currently the stuff that I could spot was:

- removing box-sizing setting to border-box
- removing setting the body margin to 0
- removing setting the font family to sans-serif

I wonder if you can share with us a more detailed list of the removed features. For now I think I am sticking the projects I am working on to version 5.0 and I am sure that the frameworks that depends on Normalize will do the same.

thierryk:
> It's 2017 everybody is setting their box sizing to border-box

Nope!

> Leaving the body margin?

Why not using a reset for that? I don't think Normalize should be the only styles sheet devs should include as their base. 

> removing setting the font family to sans-serif

If we look at Normalize as a sheet that addresses differences *between User Agent styles sheets* then it becomes easier for devs to understand **what to expect from it**.

russmaxdesign:
Totally agree with @thierryk. The aim is to set a "level playing field" only. If normalize works well, developers should not have to override anything in the normalize file. This leaves developers free to define their own base styles - including font family, margins etc - for themselves.

Toddses:
I know this is a little off-topic, @thierryk but do you have a preferred reset you use along with normalize?

thierryk:
@Toddses I'm the wrong person to ask as I'm *extremely* picky when it comes to CSS. For example, when working at Yahoo!, I removed from Normalize all the rules I considered irrelevant to our project. We did not onboard Normalize "as-is" because we were following [this approach](http://www.cssmojo.com/css-reset_versus_css-starter/). 

For what it is worth, I think we should not have to un-style things, hence why I'm not a big fan of using the universal selector (`*`) in base styles sheets.

thierryk:
Here is another reason why Normalize *should not* reset `margin` on `body`:

If we think "Mobile First" then `margin:0` on `body` may not be the best way to go. I believe such styling starts to make sense on "large" viewports. In my opinion, a page that only loads Normalize should look good/same across browsers *and* devices&mdash;and that would not be the case if we reset the `margin` on `body`.

[...]

PR #672, opened 2017-05-02
v7.0 concept: Return opinopnated rules, introduce basic.css and opinionate.css

bes-internal:
This concept of 7.0 if @necolas insists on its and return initial vision of the project for backward compatibility.
P.S. If you could make a vote, then I would vote for strict version.

see #664


necolas:
See https://github.com/necolas/normalize.css/pull/673#issuecomment-298708828

PR #673, opened 2017-05-02
v6.1 concept: Introduce opinionate.css

bes-internal:
This concept of 6.1 if @necolas decides to make the project an industrial standard. And the industry likes "use strict"
P.S. If you could make a vote, then I would vote for strict version.

see #664


necolas:
Thanks, I think a simpler solution is to revert the problematic changes in v6

PR #676, opened 2017-05-22
Update normalize.css

mattbrundage:
Update comment to reflect that certain semantic elements are supported in IE 9. According to the [Internet Explorer 9 Guide for Developers](https://msdn.microsoft.com/en-us/ie/hh410106#_HTML5_Semantic_Elements), "_By default, the elements are now styled as specified in the HTML5 specification._"

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -27,7 +27,7 @@ body {
 }
 
 /**
- * Add the correct display in IE 9-.
+ * Add the correct display in IE 8-.
  */
 
 article,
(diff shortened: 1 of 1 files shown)


Florian-R:
`IE9-` means that IE9 is out of the range, so the actual comments are correct. I agree this is confusing though.

simonsmith:
I would read that as IE 9 and below too. Maybe it could be `lte IE8`, similar to the conditional comments in HTML.

Florian-R:
After some digging in the history, it seems I was wrong, see 0c90242399c09af53850d52e992c977a538a8424, I must have confuse with another project.

Dismiss my previous comment and my apologies for the confusion.

jonathantneal:
Thanks @mattbrundage, this is an error in the documentation, and I hope this is merged.

PR #677, opened 2017-05-23
Remove unnecessary zero chars

iamtamasdohany:
Minification may solve this (?)

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -46,7 +46,7 @@ section {
 
 h1 {
   font-size: 2em;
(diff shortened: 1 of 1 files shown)


iamtamasdohany:
1 pending check

simonsmith:
As you rightly pointed out [a minifier](http://cssnano.co/) would change the values in production. I'd expect this change is unnecessary.

necolas:
Yeah, this is the job of a minifier

Issue #680, opened 2017-05-25
Fold Google Fonts stylesheets into one


csswizardry:
Currently we’re linking to two different Google Fonts stylesheets:

```html
<link href='https://fonts.googleapis.com/css?family=Source+Sans+Pro:700' rel='stylesheet' type='text/css'>
<link href='https://fonts.googleapis.com/css?family=Open+Sans' rel='stylesheet' type='text/css'>
```

We can actually fold this into just one CSS file:

```diff
-<link href='https://fonts.googleapis.com/css?family=Source+Sans+Pro:700' rel='stylesheet' type='text/css'>
-<link href='https://fonts.googleapis.com/css?family=Open+Sans' rel='stylesheet' type='text/css'>
+<link href="https://fonts.googleapis.com/css?family=Source+Sans+Pro:700|Open+Sans" rel="stylesheet" type="text/css">
```

csswizardry:
Opened a PR at #681 👍

PR #681, opened 2017-05-25
[refs #00680] Reduce Google Fonts to one stylesheet

csswizardry:
We can actually concatenate Google Fonts stylesheets, so I’ve done that
just to save us a critical request. This change only affects the
Normalize.css website, and not the Normalize.css project itself.


jonathantneal:
I like this optimization. I hope this gets merged. Thanks, @csswizardry for the PR!

Issue #683, opened 2017-05-31
Standardize font aliasing on mac os


equinusocio:
On Mac OS, webkit and mozilla render every fonts with an extra bold aliasing 

Bugzilla related issue (CLOSED): https://bugzilla.mozilla.org/show_bug.cgi?id=857142
Other references:
- http://www.dave-bond.com/blog/2009/11/fonts-too-bold-on-mac/
- https://stackoverflow.com/questions/8053389/how-to-reduce-font-weight-in-firefox-on-mac-with-css
- https://stackoverflow.com/questions/5082632/same-font-except-its-weight-seems-different-on-different-browsers
- https://css-tricks.com/forums/topic/font-appears-a-lot-bolder-on-mac-than-pc/

**The solution is to add the font smoothing options (to the `html {}`) to prevent extra bold fonts on mac os:**

```
html {
   ... 
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
}
```

On windows these properties are ignored. Note that will be a VISUAL breaking change if you add these rules to a project that is already developed on normalize.css.

PR https://github.com/necolas/normalize.css/pull/684

equinusocio:
@necolas

gouch:
I don't think you're describing this accurately. What your rules technically do is tell the browser to turn off [subpixel antialiasing](https://en.wikipedia.org/wiki/Subpixel_rendering) (a technique that improves text rendering on standard resolution LCD displays) in favor of grayscale-only antialiasing.

So _standardize_ is not the right verb here, since you're changing the default browser rendering for _every_ macOS browser to your grayscale preference. (Windows browsers also default to subpixel AA, as far I know.)

I think what you're suggesting is that the thinner grayscale AA on macOS looks closer to subpixel AA on Windows when compared side-by-side. Which may or may not be true. But what is true is that this rule would make text on a site looks _slightly off_ compared to every other site a user browses. (And compared to the text everywhere else in the browser, like menus and tabs).

This is a _*-box-sizing-border-box_ kind of change in terms of opinion and impact.

equinusocio:
> Normalize.css makes browsers render all elements more consistently and in line with modern standards. It precisely targets only the styles that need normalizing.

This way you can see fonts in the same way on each browser and on each display.

simonsmith:
> this rule would make text on a site looks slightly off compared to every other site a user browses. 
  (And compared to the text everywhere else in their browser, like menus and tabs).

[...]

PR #684, opened 2017-05-31
Standardize font aliasing on mac os

equinusocio:
Ref https://github.com/necolas/normalize.css/issues/683


equinusocio:
@necolas

necolas:
Commented on the associated issue

PR #688, opened 2017-06-11
Revert the change to remove font-family from 'html'

yuheiy:
From the README:

> Corrects bugs and common browser inconsistencies.

Should not we also revert this change?

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -4,15 +4,17 @@
    ========================================================================== */
 
 /**
- * 1. Correct the line height in all browsers.
- * 2. Prevent adjustments of font size after orientation changes in
+ * 1. Change the default font family in all browsers (opinionated).
+ * 2. Correct the line height in all browsers.
+ * 3. Prevent adjustments of font size after orientation changes in
  *    IE on Windows Phone and in iOS.
  */
 
 html {
-  line-height: 1.15; /* 1 */
-  -ms-text-size-adjust: 100%; /* 2 */
-  -webkit-text-size-adjust: 100%; /* 2 */
+  font-family: sans-serif; /* 1 */
+  line-height: 1.15; /* 2 */
+  -ms-text-size-adjust: 100%; /* 3 */
+  -webkit-text-size-adjust: 100%; /* 3 */
 }
 
 /* Sections


garrettw:
But what is the most common default for that property? Probably not `sans-serif`, I would guess.
If it were to be added back in, I'd think it should merely reflect the most common default.

yuheiy:
I think it is `sans-serif`. These are computed values for each browser.

Chrome: `sans-serif`
Firefox: `sans-serif`
Safari: `-webkit-standard`
IE: `Times New Roman`

And many websites use `sans-serif`.

`body { margin: 0 }` is not the common default of the browsers. But it is declared to remove styles that are not useful.

jonathantneal:
@yuheiy, 1. how are you determining the most widely used font on the web, and 2. would you please provide the test you are using to determine those computed values?

garrettw:
For the record, my personal installation of Firefox has never had its default font changed in Settings, and it shows a value of Times New Roman.
Further, my installation of Chrome has also not been changed, and its Standard Font is also set to Times New Roman.

yuheiy:
1. Sorry, this opinion was not grounded and it was based on my experience.
2. `getComputedStyle(document.documentElement).fontFamily` on MacOS. But my language setting is Japanese. Is `serif` selected in English?

Toddses:
In English `getComputedStyle(document.documentElement).fontFamily` is "Times" (at least in Chrome).

garrettw:
I just checked on my Windows version of Chrome by navigating to `about:blank` followed by `javascript:alert(getComputedStyle(document.documentElement).fontFamily);` and the popup said "Times New Roman".

simonsmith:
Isn't this PR more about reverting to [what was there pre `6.0.0`](https://github.com/necolas/normalize.css/blob/5.0.0/normalize.css#L14) rather than what browsers use as a default? Seeing as the other changes were reverted I can see the sense in this change for consistency sake.

jonathantneal:
@simonsmith, that’s a valid reason for restoring it, but folks are probably chiming in because the original comment left it ambiguous:

> Corrects bugs and common browser inconsistencies

And, at the time, `sans-serif` was an opinionated change; neither an inconsistency or a bug. However, @yuheiy is saying their Japanese Mac desktop defaults to `sans-serif`. That seems worth investigating.

But back to your point, @simonsmith; @necolas, if you add this back in another release for consistency sake, that leaves any research moot. To that end, I couldn’t justify the time to look into this, unless I knew the project was actually inspecting this PR as a browser inconsistency issue.

yuheiy:
Sorry, my comment was not accurate.
I navigated major browsers to `about:blank` and execute `getComputedStyle(document.documentElement).fontFamily`. Then, results are as follows.

Chrome: `"Hiragino Kaku Gothic ProN"`
Firefox: `"sans-serif"`
Safari: `-webkit-standard`

"Hiragino Kaku Gothic ProN" is Japanese sans serif.

jonathantneal:
Thanks for that update, @yuheiy. It seems then like this PR should be closed. Can you close it on your end?

yuheiy:
Bye

PR #693, opened 2017-08-13
Restore gaps in links underline in iOS 8+ and Safari 8+

antons:
#573 introduced the change. I propose to revert it, because:

1. It breaks iOS and macOS platform conventions: underlines are always drawn with “ink” style in native text views.
2. It’s a step back for web typography. Apple intentionally set the browser default to “ink” to make underlined text more readable.
3. The lack of this rule doesn’t affect other browsers, but its presence is harmful.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -96,13 +96,11 @@ pre {
    ========================================================================== */
 
 /**
- * 1. Remove the gray background on active links in IE 10.
- * 2. Remove gaps in links underline in iOS 8+ and Safari 8+.
+ * Remove the gray background on active links in IE 10.
  */
 
 a {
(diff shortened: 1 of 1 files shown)


jonathantneal:
@antons, thanks for this PR. I hope it gets merged! I think we should always fall forward.

Link to the draft spec: https://drafts.csswg.org/css-text-decor-4/#text-decoration-skip-ink-property
Link to Chrome status: https://www.chromestatus.com/feature/5631679087509504

necolas:
Merged ff155ffecdab5d04c420b6c219c974adb34492e2

antons:
Thanks! Happy to see this merged.

PR #697, opened 2017-09-01
Remove button border-radius in Chrome 62+ on macOS.

adamduncan:
Chrome 62 buttons have been restyled to resemble macOS buttons:
https://www.chromestatus.com/features/5743649186906112

Other browser vendors haven’t signalled interest.

This leads to cross-browser inconsistencies in existing UIs.

This patch matches developer expectations of buttons not having `border-radius` set by user-agents.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -272,6 +272,7 @@ select { /* 1 */
  * 1. Prevent a WebKit bug where (2) destroys native `audio` and `video`
  *    controls in Android 4.
  * 2. Correct the inability to style clickable types in iOS and Safari.
+ * 3. Remove the border-radius in Chrome 62+ on macOS.
  */
 
 button,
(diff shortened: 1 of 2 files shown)


jonathantneal:
Does this change also impact `border-width`? Per https://bugs.chromium.org/p/chromium/issues/detail?id=546679#c13

adamduncan:
Ah yes, that's right. Height of the button will remain the same, but an extra pixel given to border instead of padding?

How would Normalize intend on handling cosmetic differences like that, as opposed to border-radius? I feel border radius would have more unexpected consequences.

adamduncan:
Closing as Chrome reverting change. See: https://github.com/necolas/normalize.css/issues/706#issuecomment-342108033

cannikin:
Might want to re-open since it looks like Chrome actually did go live with this. See https://github.com/necolas/normalize.css/issues/706#issuecomment-344995038

mgol:
This change landed in Chrome 62 but will be reverted in Chrome 63 which will arrive around Dec 5 (see https://www.chromium.org/developers/calendar).

mgol:
Chrome 63 has just been released.

PR #698, opened 2017-09-06
Inherit font family for inputs

Sky161:


jonathantneal:
This seems like a better opinionated default, @Sky161, and I’m generally against opinionated styles in normalize. 😄 If normalize wants to keep opinionated changes, I would prefer this one.

necolas:
Merged ca8a357baf6e724ab84843ac4607f8d2024cf0f8

PR #701, opened 2017-09-16
Add support for Eyeglass

### LINKED ISSUE #638: Any plans to make Normalize eyeglass compatible?
I was just wondering if there was any plan to make the NPM module of Normalize compatible with [Eyeglass](https://github.com/sass-eyeglass/eyeglass)?

I’ve seen there’s a [sass version of normalize](https://www.npmjs.com/package/normalize-scss) which is eyeglass compatible, but all I really want to pull into my projects is the straight up Normalize CSS, so was wondering if this was planned, or whether it would be worth me forking the repo and making a module that does this.  Alternatively, I’d be happy to submit a PR to add this, if it would be something that you’d want to look to include.

Thanks for all the great work on the project.

Snugug:
#638 was closed by saying "This could have a PR for it…", so I made a PR for it!

Quick explanation of what's happening:
- Add `eyeglass` fields to `package.json`
- Add `keyword` field to `package.json` and include the `eyeglass-module` keyword
- Update the `files` field in `package.json` to include the requisite `eyeglass-exports.js` file and the `to-sass.js` file (more on that in a moment)
- Add the `eyeglass-exports.js` file to tell Eyeglass where the files are stored
- Add `to-sass.js` file to convert `normalize.css` in to `_normalize.scss` after NPM install
- Add `_normalize.scss` to git ignore

Let me talk through `to-sass.js` because it's likely the most controversial of the additions. Sass cannot import raw CSS files, like what is currently distributed. As such, in order to have full compatibility with Eyeglass as expected for Sass authors, we need a Sass file. This is simple enough, change the `.css` extension to `.scss` and it can be imported (I also rename it to a Sass partial so it doesn't compile its own file, by prepending `_` to the file name). This could be a 1-liner with `cp normalize.css _normalize.scss`, but then `postintall` will fail on terminals that don't have a Bash-compatible `cp` function available. So, I needed to write a tiny file to A) delete `_normalize.scss` if it's already there and B) replace it with a clean copy based off of `normalize.css` making sure it's up-to-date. All of this is done with native Node libraries that are available back to version `0.12` of Node, meaning this should work basically everywhere that Node works, and absolutely every version of Node that is not [End of Life'd](https://github.com/nodejs/LTS)

I have tested installing and using Normalize from my branch in a project I already have set up. Presuming Eyeglass is already set up, using it in a Sass file after running `npm install normalize.css --save-dev` is `@import 'normalize'`.

---

Resolves #638

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules
+_normalize.scss
--- a/eyeglass-exports.js
+++ b/eyeglass-exports.js
@@ -1,0 +1,7 @@
+var path = require('path');
+
+module.exports = function(eyeglass, sass) {
+  return {
+    sassDir: path.join(__dirname)
+  }
+}
--- a/package.json
+++ b/package.json
@@ -6,15 +6,26 @@
   "style": "normalize.css",
   "files": [
     "LICENSE.md",
-    "normalize.css"
+    "normalize.css",
+    "eyeglass-exports.js",
+    "to-sass.js"
   ],
+  "eyeglass": {
+    "needs": "^1.0.0",
+    "exports": "eyeglass-exports.js",
+    "name": "normalize"
+  },
   "devDependencies": {
     "stylelint": "^7.9.0",
     "stylelint-config-standard": "^16.0.0"
   },
   "scripts": {
-    "test": "stylelint normalize.css"
+    "test": "stylelint normalize.css",
+    "postinstall": "node ./to-sass.js"
   },
+  "keywords": [
+    "eyeglass-module"
+  ],
   "repository": "necolas/normalize.css",
   "license": "MIT",
   "bugs": "https://github.com/necolas/normalize.css/issues",
--- a/to-sass.js
+++ b/to-sass.js
@@ -1,0 +1,21 @@
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+
+const pkg = require('./package.json');
+
+// Output file paths
+var input = path.join(__dirname, pkg.main);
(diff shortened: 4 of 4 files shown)


Toddses:
Sass _can_ import raw CSS files. We import normalize.css on every project just fine.

Snugug:
@Toddses While I have no doubt your setup is able to import CSS files to Sass just fine, it is [absolutely](https://github.com/sass/sass/issues/556) [not](https://github.com/sass/libsass/pull/1963) [standard](https://github.com/sass/libsass/pull/1964) [behavior](https://github.com/sass/libsass/issues/2096) for any of the 4 major Sass implementations (Dart Sass, Ruby Sass, libsass, or Node Sass). Even if it were, this PR is specifically addressing an issue that was opened about adding Eyeglass support, which according to that issue the authors weren't against but wanted a PR for. Here's that PR

Snugug:
(in fact, I wrote [node-sass-import-once](https://github.com/at-import/node-sass-import-once) specifically to address CSS not being importable in Sass as a default behavior)

Toddses:
Our setup isn't anything special, pretty common I think. Based on my own searching, libsass supports `@import` of CSS files as a partial as of a couple years ago (https://github.com/sass/libsass/pull/754). Ruby Sass is promising to support this out-of-the-box in 4.0 (https://github.com/sass/sass/issues/193).

I have no sway over this project, I was just commenting on the "controversial" change as a user.

jonathantneal:
I think the original request was limited to updating package.json to better accommodate eyeglass, and this is overkill. I would not want this merged.

necolas:
Thanks but this project has never catered to specific tools

PR #709, opened 2017-11-13
Update normalize.css

vanthome:
The same rule selector `b, strong` is used twice which causes problems in some scenarios and does not match the convention in the rest of the code. I have merged both rules under one selector.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -117,21 +117,14 @@ abbr[title] {
 }
 
 /**
- * Prevent the duplicate application of `bolder` by the next rule in Safari 6.
+ * 1. Prevent the duplicate application of `bolder` by the next rule in Safari 6.
+ * 2. Add the correct font weight in Chrome, Edge, and Safari.
  */
 
 b,
 strong {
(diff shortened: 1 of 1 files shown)


jonathantneal:
Hey @vanthome, thanks for looking through the code of this project. Did you see the comment next to the line in question?

> Prevent the duplicate application of `bolder` by the next rule in Safari 6.

This duplication is purposeful.

You did not specify *problems in some scenarios*, so could you be more specific about this issue?

You also said this *does not match the convention in the rest of the code*. There are other parts of the code that do similar things, and I assure you we get many issues and PRs for the duplicate [`monospace, monospace`](https://github.com/necolas/normalize.css/blob/master/normalize.css#L91). 😄 

Again, it’s awesome that you are looking into the code, and that you would take the time to make a clean PR that only affects the parts of the code you want to change.

My two cents; do not merge.

vanthome:
@jonathantneal ok, now I understand the comment and code. I withdraw the PR.

PR #710, opened 2017-11-15
Reset border-radius on button element

haniotis:
Overwrites https://www.chromestatus.com/features/5743649186906112

@necolas


garrettw:
See #697. This was already proposed but closed because Chrome decided to revert the change.

haniotis:
🎉

Issue #632, opened 2016-10-04
normalize.min not in bower


ferreiro:
Hello guys,

I've just downloaded normalize.css using bower. But in bower_components/normalize.css does not appear the minified version. Is this because is not possible to include it on bower or because you don't want to suppor the minied version?

Best.
Jorge.

Toddses:
#410

jonathantneal:
What’s mentioned in #410 (thanks @Toddses), and it’s much cleaner for us all to leave compiled files **off** the repository. If you’re using bower or npm to import normalize, I would imagine you have a build process that could combine and minify all these files for you.

ferreiro:
@Toddses Thank you!
@jonathantneal Ah okey! I use gulp for build tasks. So according to you're response I'll then create a new pipeline to minify the normalize for production. Thanks for your response ;)

Toddses:
@ferreiro If you're using Sass in that build process, you can just `@import` normalize straight from your node modules and it will become a part of your core css file.

ferreiro:
@Toddses Thanks for your response! You mean to do @include 'test.css' in the main file for the sass? If this is the scenario, I've just tryed it and does not work.

Toddses:
@ferreiro You need to `@import "test.css"`. Check out the Import section here http://sass-lang.com/guide

josephfusco:
I might be out of the loop with recent versions of sass but as far as I'm aware using `@import` on a regular .css file does not treat it like a partial but rather how regular CSS does.

Toddses:
I'm using gulp-sass and `@import`ing normalize.css and it's treating it like a partial. But I did make a mistake above, you'd want to `@import "test";`. No extension.

Issue #352, opened 2014-07-22
Can we have minified copies in Releases (not Repo)?


whz-pa:
The 'Releases' feature, adds support for hosting builds and binaries of a version (manually or tag-driven).
So, is there any plan for adding a minified version during a release?

p4bloch:
To provide an actual answer to this question instead of making assumptions on everyone else's setup, the CDN already provides a minified version for this library
https://cdnjs.com/libraries/normalize

Issue #338, opened 2014-06-02
Minified Version


ajohnson6494:
Could you please offer a minified version of this code so that it can be used in production sites. Thanks!

PR #714, opened 2017-11-20
Added minify support

ardagedikk:
I think it would be better to present the reduced size.
Size before shrinking 7.27kb, size after shrinking 2.28kb


jonathantneal:
Hi @ardagedikk, I appreciate your consideration for the bytes. It may have saved you some time to check the issue tracker. It was resolved to let the end-user minify the library.

> Going with this: people can minify using that online YUI Compressor-based tool (included in the readme), or as part of their build process.
> — https://github.com/necolas/normalize.css/issues/4#issuecomment-1783421 (Aug 11, 2011)

Don’t feel too bad. In the last six years, it comes up again from time to time. See #4, #75, #81, #116, #131, #146, #224, #338, #352, #382, #410, #632.

Finally, some good news — you can still get what you want. The recommended CDN already provides a minified version for this library. See https://cdnjs.com/libraries/normalize

ardagedikk:
It could be a good solution for those who don't want to use CDN and don't want to deal with minifying

garrettw:
Why wouldn't you want to use CDNjs? It's hosted -- you don't have to pay for your own CDN. It's literally just as easy as linking to your own CSS file on your website. I can't think of a good reason not to use it.

ardagedikk:
I am not against using CDN, but I think the use of CDN in projects where there is a lot of dependency is a disadvantage in managing dependencies. In such cases it is logical to install the package with the package manager and use the minify state.

jonathantneal:
I still believe this should be closed. Again, this has been addressed in the issue tracker.

necolas:
Yeah this has been discussed before. Thanks but I won't be merging

Issue #719, opened 2017-12-19
Installing this broke my webpack react config


yongelee:
Basically my webpack config was running smoothly until I decided to npm install this package. All I did was shut down my dev server, npm install this, then run it again. THen everything broke.

I tried deleting the folder and removing it from my package.json but still broke everything.

I actually can't believe my perfect webpack config with hot reloading and react router broke from just installing this package. Wow it's insane how much a poorly designed module can ruin your whole project.

ERROR in multi (webpack)-dev-server/client?http://localhost:8080 webpack/hot/dev-server babel-polyfill react-hot-loader/patch webpack-dev-server/client?http://localhost:8080 webpack/hot/only-dev-server ./src/index.js
Module not found: Error: Can't resolve 'babel-loader' in 'C:\Users\Albert\Documents\dotahub\client'
 @ multi (webpack)-dev-server/client?http://localhost:8080 webpack/hot/dev-server babel-polyfill react-hot-loader/patch webpack-dev-server/client?http://localhost:8080 webpack/hot/only-dev-server ./src/index.js

ERROR in multi (webpack)-dev-server/client?http://localhost:8080 webpack/hot/dev-server babel-polyfill react-hot-loader/patch webpack-dev-server/client?http://localhost:8080 webpack/hot/only-dev-server ./src/index.js
Module not found: Error: Can't resolve 'react-hot-loader/patch' in 'C:\Users\Albert\Documents\dotahub\client'
 @ multi (webpack)-dev-server/client?http://localhost:8080 webpack/hot/dev-server babel-polyfill react-hot-loader/patch webpack-dev-server/client?http://localhost:8080 webpack/hot/only-dev-server ./src/index.js

jonathantneal:
> Wow it's insane how much a poorly designed module can ruin your whole project.

I’m really sorry you’re having a rough day. I’ve had rough days. They suck. So, have you checked if `babel-loader` is installed?

I imagine that after you uninstalled `normalize.css`, you hopefully began to suspect that this issue had nothing to do with normalize.css. You probably [checked the `scripts` field in normalize.css’s package.json](https://github.com/necolas/normalize.css/blob/master/package.json#L15-L17) just to be sure normalize.css doesn’t _do_ anything.

After you reviewed your error logs, noticed that issue with `babel-loader`, and reflected on how normalize.css really is _just_ a css file, as the project name suggests — I get it — you wanted to let off some steam. After all, you thought installing a module was all that changed.

[...]

PR #720, opened 2017-12-19
Fixed repository line in package.json

garrettw:
It probably doesn't relate to #719, but I did notice this one thing that looked off, so I thought I'd fix it just in case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
   },
   "repository": {
     "url": "https://github.com/necolas/normalize.css",
-    "type": "git
+    "type": "git"
   },
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


jonathantneal:
Hello. This doesn’t fix anything. See https://docs.npmjs.com/files/package.json#repository

> For GitHub, GitHub gist, Bitbucket, or GitLab repositories you can use the same shortcut syntax you use for npm install.

garrettw:
😞

jonathantneal:
Thanks for offering a contribution. No need to be sad. You are just trying to help. 😄

garrettw:
I know, I just feel a bit stupid, that's all 😝

PR #730, opened 2018-02-15
Render the main element consistently in IE

### LINKED ISSUE #727: `main` has been removed but is needed for IE 11
```css
main {
    display: block;
}
```
is needed for IE 11 to render the element properly. Probably also for IE 10.

thasmo:
Fixes #727 by rendering the main element as block-element for IE.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -25,6 +25,14 @@ body {
 }
 
 /**
+ * Render the `main` element consistently in IE.
+ */
+
+main {
+  display: block;
+}
+
+/**
  * Correct the font size and margin on `h1` elements within `section` and
  * `article` contexts in Chrome, Firefox, and Safari.
  */


jonathantneal:
@thasmo, nice, but would you specify the target browser(s)?

thasmo:
Sure, where exactly? In the PR title, commit message, comment? :)

jonathantneal:
Line 28, where you specified **all browsers*. https://github.com/necolas/normalize.css/pull/730/commits/55f0ce1a29a6669de43ec80096f7ce88bb067688#diff-bb3dde41d97f19be8ab7b4780a915d5eR28

Before it was removed, I had specified `Add the correct display in IE` here:

https://github.com/necolas/normalize.css/blob/6.0.0/normalize.css#L49

thasmo:
@jonathantneal, done. 👍

thasmo:
Oh, I see. On it.

jonathantneal:
Just putting `IE` will do, because that means all versions. I hope other IE11 styles weren’t lost, unless this project no longer supports IE11 either.

jonathantneal:
I know it sounds really picky, but `IE 9-11` would not be accurate, because IE8 did not support `main` 🤷‍♂️  and this project doesn’t support IE9 or IE10 AFAIK.

thasmo:
@jonathantneal, last chance. :P

thasmo:
Actually, last chance now. ;)

thasmo:
Typed enough Is and Es today. 👍

thasmo:
I believe IE 10 is supported, at least that's what's documented but I'd be happy to only have support for IE 11.

jonathantneal:
Yea, it’s not complicated. The comment just needs to blanket IE.

It will also need to use the same voice as the other comments, which would be:

```
Add the correct display in IE
```

That’s not the message you’ve written, as you’ve introduced new terms like `Render` and `consistently`. As picky as it sounds, you’d likely lose that one line in git blame 😄 but otherwise this looks great and I hope it’s merged.

I really have zero control over this. I’m just looking out for the project as best I can.

nzec (approved):
Looks good! 👍

fritzmg:
👍 this caused some headache for me in a current project where the `<main>` element was used and the `padding` for example had no effect in IE. I didn't realize IE 10+ does not treat `<main>` as a block level element automatically.

thasmo:
🤔

Tetonne:
get standard scrollbar (safari)

garrettw:
@thasmo @Tetonne please do not post meaningless/spam comments.

fritzmg:
Geez, _exactly_ 3 months later I stumbled over the same issue - at least it only cost me half an hour this time, until I figured it out... __again__ 😁 

Why has this not been merged and released in 8.x?

GarethPW (approved):

fritzmg:
@necolas the NPM package still only has 8.0.0: https://www.npmjs.com/package/normalize.css?activeTab=versions
Could you may be check why?

necolas:
Forgot to publish. Should be up now. Thanks

nzec:
@necolas Thanks! ❤️

PR #735, opened 2018-03-27
form elements inherit line height from ancestor

jzilg:
I think it is more practical if the form elements inherit the line height instead of set it fix to 1.15. So if you change the line height of your root or any ancestor element, it will apply to your form elements.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -156,7 +156,7 @@ select,
 textarea {
   font-family: inherit; /* 1 */
   font-size: 100%; /* 1 */
-  line-height: 1.15; /* 1 */
+  line-height: inherit; /* 1 */
   margin: 0; /* 2 */
 }
 


frenzzy:
Maybe we can do the same with `font-size` ?

safareli:
big 👍  for inheriting font-size and line-height

benface:
> inheriting the value might end up using a different computed value if the parent is using a em or percentage for the font-size, as the inherited value is not the computed one, but the percentage itself

I just tested this and found it to be incorrect. There is no difference between `inherit`, `100%`, and `1em` even if the parent has a `font-size` in `em` or `%`.

J-Vin:
Is this still open? I do not know what this is calling for as a fix, but it does read like a possible solution. https://runkit.com/j-vin/5f3a19941c8c0a0019d43469

PR #744, opened 2018-04-26
inherit line-height for form elements

safareli:
`line-height` is already set to `1.15` on `html`, so we can `inherit` the value, plus if `line-height` is overridden it will be cascaded.


safareli:
closing in favor of #735

Issue #151, opened 2012-12-23
Normalize `select` element styling


axgmc:
To allow `select` to have a height on Safari for Mac, and add some consistency between browsers, just add:

``` css
@media screen and (-webkit-min-device-pixel-ratio:0) { /* For Safari and Chrome */
    select {
        border-color: #adadb4; /* Turns off Aqua */ 
    }
}
```

Yapaslfeu:
Another way far more consistent to deactivate Aqua is to touch the border-image property:
select {
	border-image-source: url();
}

PR #746, opened 2018-05-07
Remove the default apparence of select on all browsers

vretamal:
Avoid select' styles issues if you're develop on Windows or OSX.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -181,6 +181,15 @@ select { /* 1 */
 }
 
 /**
+ * Remove the default apparence of select on all browsers
+ */
+select {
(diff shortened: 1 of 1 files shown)


garrettw:
Based on @necolas' comment on [this issue](https://github.com/necolas/normalize.css/issues/151#issuecomment-32418209) which he tagged as "won't fix", I don't think that's going to happen.

vretamal:
you're right. I've searched if someone have tried the fix and I didn't find any.
Well, I must wait for his opinion.
Grettings
(sorry my bad english)

necolas:
No thanks for reasons previously explained

PR #754, opened 2018-07-10
border-bottom: 0 instead of none

markgoho-EDT:
I guess it's somewhat subjective, but some folks prefers the `0` over `none`. Take it or leave it.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -75,7 +75,7 @@ a {
  */
 
 abbr[title] {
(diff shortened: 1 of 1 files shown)


garrettw:
Just to add info to the issue: I found [this codepen](https://codepen.io/denilsonsa/pen/LkdHh?editors=110) which explains the differences in various situations. Hopefully it will help @necolas determine which way is best.

PR #758, opened 2018-08-18
Add html5 display-role reset for main element

MadeByMike:
Old versions of IE once required a bunch of html5 elements display to be set to block. The browsers that required this are possibly out of scope, however, since the `main` element was added later, a reset is still required for this element in IE11. See: https://github.com/necolas/normalize.css/issues/727

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -339,3 +339,11 @@ template {
 [hidden] {
   display: none;
 }
+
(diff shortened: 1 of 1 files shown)


necolas:
df07c00a92c683e591b68bcc45075ff7253b7482

Issue #737, opened 2018-03-29
Add global border-box


JonathanSchndr:
i only coding with border-box, maybe a enhancement?

```
*, *:before, *:after {
  box-sizing: border-box;
}
```

source: https://css-tricks.com/box-sizing/

miquelfire:
Outside the scope of the project.

PR #761, opened 2018-08-24
Unify box-sizing

axelrindle:
The `box-sizing` property has different defaults in different browsers, so this should be unified. The border should be included by default, so the normalized value should be `border-box`.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -4,6 +4,14 @@
    ========================================================================== */
 
 /**
+ * Unify the box sizing to include borders.
+ */
+
+* {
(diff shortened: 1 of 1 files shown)


garrettw:
See: #202 #409 #737

axelrindle:
@garrettw Ok, got it. Thanks for clarifying 👍

garrettw:
@axelrindle I do have a project of my own creation that includes a slightly different normalize than this one, and it does include `box-sizing`. https://github.com/garrettw/vitals

PR #763, opened 2018-09-13
Correct the width and height on 'body' element in all browsers

nazarepiedady:


mvasilkov:
I would assume this is out of scope for normalize.css (also why set width at all).

garrettw:
Which browsers need this to match the rest?

mvasilkov:
@garrettw it's not the default in any browser I know of.

nazarepiedady:
sorry guys, i think that could be a good features for normalize.css, because i had some problems with this, i want to add this properties to optimize the job.

garrettw:
@nazarepiedady The point of normalize.css is only to fix browser behavior that is outside the norm established by other browsers. They have moved away from these so-called "opinionated" styles because, while they might be nice to have, they are not in line with this project's goals.

nazarepiedady:
sorry, I really sorry!.

necolas:
No thanks. This changes the default `body` layout. That's something you can do in your own code if you want.

PR #766, opened 2018-10-01
Add min css file

THE-BLACKSHEEP:
For faster than version


Toddses:
Not going to happen, per a number of previous discussions. #4 for example.

PR #767, opened 2018-10-18
Text decoration and box sizing webkit and moz support

Richienb:
In this PR, support for the `text-decoration` and `box-sizing` are now available for `webkit` and `moz`.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -44,8 +44,8 @@ h1 {
 
 hr {
   -webkit-box-sizing: content-box;
-     -moz-box-sizing: content-box;
-          box-sizing: content-box; /* 1 */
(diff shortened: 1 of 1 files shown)


monkeyhybrid:
`box-sizing` does not need any prefixes for any [officially supported browser](https://github.com/necolas/normalize.css#browser-support).

I'm not sure `text-decoration` needs any either, even for your `dotted` value that comes from newer, `text-decoration-style` shorthand. [Browsers appear to either support it, or not](https://caniuse.com/#search=text-decoration-style), regardless of the prefix.

necolas:
This isn't necessary

Richienb:
Ok.

PR #769, opened 2018-11-14
Maded line-height bigger

IgnaciodeNuevo:
Added a bigger line-height helps allways to read better the text, so users have no problem. Shouln't the minimum recomended be bigger than 1.15? I always find as typography lover less than 1.25 to small.

Also recently read [Webfont Handbook](https://abookapart.com/products/webfont-handbook), [On Web Typography](https://abookapart.com/products/on-web-typography), [Flexible Typesetting](https://abookapart.com/products/flexible-typesetting) and [Better Web Type](https://betterwebtype.com/) and they recommend bigger line-height.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -9,7 +9,7 @@
  */
 
 html {
-  line-height: 1.15; /* 1 */
+  line-height: 1.4; /* 1 */
   -webkit-text-size-adjust: 100%; /* 2 */
 }
 


garrettw:
Normalize.css is intended only to give all browsers the same starting point in CSS to make development easier. If most browsers use a default `line-height` of 1.15, then it should be set to 1.15 here to fix the rest of the browsers to be the same. Further changes to `line-height` should be made either in other libraries or your own CSS.

tomasz1986:
> If most browsers use a default `line-height` of 1.15, then it should be set to 1.15 here to fix the rest of the browsers to be the same.

He does have a point though, as this particular `line-height` was likely adopted by the browsers in the times when the pinnacle of screen resolution was 640x480 (and whole World Wide Web looked much different than today). Nothing can be done about it though, unless the browser vendors themselves decide to change their defaults.

garrettw:
I don't disagree with the sentiment -- it just shouldn't be done like that in this library at this time.

hd4ng:
It also set line-height: 1.15 for the button, input, optgroup, select, textarea
`
button, input, optgroup, select, textarea {
  font-family: inherit;
  font-size: 100%; 
  line-height: 1.15;
  margin: 0; 
}
`
We should also change the value here

Issue #736, opened 2018-03-28
iOS placeholder line-height with some fonts


denis-sokolov:
On iOS 10, with certain fonts, the vertical alignment of input placeholders is broken because of a strict line-height defined in normalize.css.
https://github.com/necolas/normalize.css/blob/7369f566898b8cefc2649b6a66e112857de4c46f/normalize.css#L159

In my particular case I have a font Galano Grotesque Alt, and here is how my input looks on iOS:
![image](https://user-images.githubusercontent.com/113721/38041522-a09eb624-32ba-11e8-83c7-05f532f68c93.png)

If I change the line-height to `normal`, iOS fixes the layout:
![image](https://user-images.githubusercontent.com/113721/38041549-a9caee02-32ba-11e8-94a3-171334232f69.png)

I can’t make an interactive example because the font is not freely available.

This only happens with the placeholder, only on iOS, only with this font in my case. If I start typing in, it aligns correctly:
![image](https://user-images.githubusercontent.com/113721/38041702-1a6f0062-32bb-11e8-9cef-022bbec1f06b.png)

A general problem with vertical alignment of the placeholder in iOS seems to be prevalent. A web search finds numerous relevant questions [on Stack Overflow](https://stackoverflow.com/a/22657686).

The line-height change was introduced in #607 and is nicely explained in [a comment on #591](
https://github.com/necolas/normalize.css/issues/591#issuecomment-230976875).

I don’t suggest to change that line-height right now. The interaction is obscure, and there seems to be value in the value of line-height as is. I want to preserve this research in case anyone else stumbles in a similar scenario and searches the issue tracker for the issue. So feel free to close this immediately as inactionable, or keep open if you want to ponder more about it.

1aron:
@denis-sokolov The following could fix your problem 👍 
```
::placeholder {
    line-height: normal;
}
```

trylovetom:
```
input::placeholder,
input::-webkit-input-placeholder,
input::-moz-placeholder {
  line-height: normal !important;
}
```

gmhenderson:
For whatever reason, I needed:

```
::placeholder {
  line-height: normal !important;
  vertical-align: middle;
}
```

Alfrex92:
I had the same problem and I fixed with 

```
input {
  -webkit-appearance: none;
  border-radius: 0;
 :focus {
    outline: none;
    border-radius: 0;
  }
  :active {
    border-radius: 0;
  }
}
```
[...]

Issue #587, opened 2016-04-17
Try making forms looking the same once for all!


Yapaslfeu:
I know you will not agree (#578), but i believe that the only solution to get rid of all those inputs style problems we have is to completely remove all browsers specific UI style, so all forms will look exactly the same on every browser!

And this can be done easily, starting from setting the same border and background color to all fields!
I know it's arbitrary colors but one can change them easily, reducing the problem to a simple design question!
The result is that we don't have to think about browsers specifics anymore and only concentrate on design, forgetting about compatibility/consistency.

Here is the main part of the code i use for that (without the focus and hover parts):
`button,`
`input:not([type="range"]):not([type="file"]):not([type="image"]),`
`select,`
`textarea {`
`	border: 1px solid #c0c0c0;`
`}`
`[type="color"] {`
`	background-color: #fff;`
`}`
`[type="file"] {`
`	background-color: transparent;`
`}`
`button,`
`html [type="button"],`
`[type="reset"],`
`[type="submit"] {`
`	background-color: #e1e1e1;`
`}`
`::-webkit-file-upload-button {`
`	border: 1px solid #c0c0c0;`
`	background-color: #e1e1e1;`
`}`
`::-ms-browse {`
`	border: 1px solid #c0c0c0;`
`}`

Yapaslfeu:
Here is a demo with the full CSS i use: [https://codepen.io/Yapaslfeu/pen/aNGLQy]

jonathantneal:
This would be much more useful in something like [sanitize.css](https://github.com/10up/sanitize.css),. normalize.css should not get into overriding OS styling.

Yeswanth-JG:
Hi. @Yapaslfeu  ...

Normalize preserves default styling. It standardizes the things and maintain the consistency. Changing the presentation is not intended.

battaglr:
@Yapaslfeu, as others already said this doesn't seem suitable for this project.

*Off-topic:* as I can see from the many issues that you have opened over time, you know a lot about forms and browsers inconsistencies; would be awesome if you created a project with all that (please, provide a link if you do)!

jonathantneal:
Agreed.

visicode:
@battaglr, thank you for your suggestion!
I just create the new project **_formalize.css_** intended to normalize forms:
https://github.com/visicode/formalize.css

It has to be included right after normalize.css.
This is a very first version and any help will be appreciated to improve it…

battaglr:
[...]

PR #775, opened 2019-03-21
clean placeholder

1aron:
1. line-height: 1.15 would make `::placeholder text` move up ( in safari ),
but `::placeholder { line-height: normal }` could fix it.

2. make `::placeholder` opacity default to `1`

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -9,8 +9,10 @@
  */
 
 html {
-  line-height: 1.15; /* 1 */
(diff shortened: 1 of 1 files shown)


Toddses:
It appears you've changed the indentation from 2 spaces to 4.

garrettw:
Please look over #550, #587, #736 -- there has been a lot of discussion about these issues, and decisions were made on how they should be handled.

1aron:
#776

PR #773, opened 2019-01-03
Combine `pre` and `code,kbd,sample` declarations

erdaltsksn:
why: Because two declaration have same properties. We don't need to
         define them twice.

issue: close #772

PR #781, opened 2019-05-01
Combine 'template' and '[hidden]', 'main' and 'details', 'button, input' and 'hr' declarations

itfarrier:
This is good for readability and size.

We have pull and issue with the same idea [https://github.com/necolas/normalize.css/pull/773](https://github.com/necolas/normalize.css/pull/773)


necolas:
No thanks

PR #782, opened 2019-05-07
chore(Comments): Mark license comment as unimportant

tomciopp:
@necolas Based on the way this comment is written, it will not be removed by default by a lot of CSS optimizers. Considering how widespread this project is and the fact that this project is MIT licensed I don't believe there is any reason why it should be preserved after minimization.


necolas:
Preserving licenses is expected and required

PR #790, opened 2019-08-09
Added style to body to improve font rendering

ivoilic:
Adding these styles to the body improves font rendering on many browsers/device dramatically.
This would be a great addition to normalize.css as it would improve the consistency of how fonts are displayed across browsers.
Here is an article for reference: https://medium.com/better-programming/improving-font-rendering-with-css-3383fc358cbc

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -22,6 +22,8 @@ html {
 
 body {
   margin: 0;
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
 }
 
 /**


mvasilkov:
Previously: https://github.com/necolas/normalize.css/issues/683

Related: http://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/

ivoilic:
@mvasilkov I think it's worth at least taking another look. It's worth considering that many of the most visited sites in the world use this: https://developer.microsoft.com/en-us/microsoft-edge/platform/usage/css/-webkit-font-smoothing/ (Data from 2017)

Here is Twitter's solution ([source](https://www.zachleat.com/web/font-smooth/)):
`@media screen and (-webkit-min-device-pixel-ratio: 2), screen and (min-resolution: 2dppx) {
	body {
		-moz-osx-font-smoothing: grayscale;
		-webkit-font-smoothing: antialiased;
	}
}`

Toddses:
Your Twitter example is less broad than what you're suggesting with this PR. It also doesn't normalize browser styles - no browser sets this by default - which is what normalize does.

ivoilic:
@Toddses Fair enough. I hadn't thought of if it like that.

PR #791, opened 2019-08-12
Added An HTML Attribute

mouadTaoussi:
Added An **HTML** Attribute Which Is 'download' Inside `<a>` Tag In `btn-download` For Allow Any  One To **Download The File Directly Instead Just Showing The Source File**

### DIFF
--- a/index.html
+++ b/index.html
@@ -22,7 +22,7 @@
 
         <div class="cta-option">
           <a class="btn-download"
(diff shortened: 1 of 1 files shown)


frenzzy:
It is a bit useless here since:
> [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/A#attr-download): This attribute only works for [same-origin URLs](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy).

PR #792, opened 2019-10-10
Changes in margin

priyanshi2811:


garrettw:
I think you may misunderstand the purpose of this repo. It is intended to bring all browsers in line with each other's default CSS. Most browsers use a body margin of `0`, not `15px`, so changing this would break a TON of websites that expect it to be `0` by default.

Also `italic` is not a font family, and browsers do not commonly give `<pre>` blocks italic text, so this would similarly be an inappropriate change.

There are other CSS libraries that are more geared toward setting opinionated defaults, so if you think your changes are an ideal starting point for styling a website, I would recommend suggesting your changes to one of those libraries.

PR #793, opened 2019-10-16
Simplify legend element normalization

drify:
See https://github.com/necolas/normalize.css/issues/100#issuecomment-197119273 and https://github.com/necolas/normalize.css/pull/536 . These two declarations are intended to fix IE8, which normalize.css dropped support long ago.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -240,10 +240,8 @@ fieldset {
 legend {
   box-sizing: border-box; /* 1 */
   color: inherit; /* 2 */
-  display: table; /* 1 */
   max-width: 100%; /* 1 */
   padding: 0; /* 3 */
-  white-space: normal; /* 1 */
 }
 
 /**


Toddses:
Both noted issues say this fix covers IE8 - 11. Normalize is still supporting IE10+.

drify:
> 
> 
> Both noted issues say this fix covers IE8 - 11. Normalize is still supporting IE10+.

Hello,

Thanks much for your reply! But please check that again.
> **IE9-11** and **Edge 12-13** can be completely fixed using `box-sizing: border-box; max-width: 100%`, but IE8 will still experience some issues as “_the min-width property applies to content-box even if box-sizing is set to border-box_” [cite](http://caniuse.com/#search=content-box).

So remove the other two lines should be fine.

drify:
@Toddses @necolas Any ideas?

nickmad1000:
```````

@@ -240,10 +240,8 @@ fieldset {
 legend {
   box-sizing: border-box; /* 1 */
   color: inherit; /* 2 */
-  display: table; /* 1 */
   max-width: 100%; /* 1 */
   padding: 0; /* 3 */
-  white-space: normal; /* 1 */
 }
 
 /**

```

```

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #795, opened 2019-10-19
body padding added

lakshmipriyamukundan:
body padding changes


Toddses:
do _any_ browsers default the `body` padding to 0?

PR #796, opened 2019-10-20
updated normalise.css

NeuralNinja57:
display inline


Toddses:
do _any_ browsers default the `main` element to `inline`?

bznwbb222 (approved):

Issue #800, opened 2019-10-26
Anchor tag has an outline when got focus on Firefox on Ubuntu


mlvnbrj:
I am using Mozilla Firefox 70.0 (64bit) on my Ubuntu 18.04 OS, I already notice it for a long time but I ignored it, until I open the test.html and I read that anchor tag `should not have a focus outline when both focused and hovered (opinionated)` _**see 6.3**_

[...]

PR #801, opened 2019-10-26
Remove outline of anchor tag when got focused on Firefox for Ubuntu

mlvnbrj:
Issue #800 

- I added `outline: 0` to fix it
- 1 file changed `normalize.css`

See my [Codepen Example](https://codepen.io/milbeeen/pen/eYYRgdM)

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -70,11 +70,13 @@ pre {
    ========================================================================== */
 
 /**
- * Remove the gray background on active links in IE 10.
+ * 1. Remove the gray background on active links in IE 10.
+ * 2. Remove outline when got focused using Firefox on Linux.
  */
 
 a {
-  background-color: transparent;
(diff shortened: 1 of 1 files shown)


clarfonthey:
This is bad for accessibility, and against the spirit of the project, which is simply to provide consistency across browsers.

See: http://www.outlinenone.com

mlvnbrj:
> This is bad for accessibility, and against the spirit of the project, which is simply to provide consistency across browsers.
> 
> See: http://www.outlinenone.com

Outline on focused state of an `a` element is also bad for design sometimes (in my opinion only just like normalize said on the [test.html](https://github.com/necolas/normalize.css/blob/master/test.html)), accessibility is better to be done by adding alternative styles.

clarfonthey:
Yes, but don't remove it without adding that alternative style.

PR #802, opened 2019-11-06
Removed The Double Declaration

mubshrx:
Removed The Double Declaration of "**text-decoration: underline dotted;** /* 2 */ "on line 87.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -84,7 +84,6 @@ a {
 
 abbr[title] {
   border-bottom: none; /* 1 */
-  text-decoration: underline; /* 2 */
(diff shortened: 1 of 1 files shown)


fisker:
https://caniuse.com/#search=text-decoration

mubshrx:
sorry! I don't understand what do you want to tell me?

Michaelsy:
> sorry! I don't understand what do you want to tell me?

Some browsers do not understand `text-decoration: underline dotted;`.
In these cases, `text-decoration: underline;` is used.
Therefore, this is not a double definition.

PR #803, opened 2019-11-14
Add a normalized 'a'

lpeil:
Remove the text decoration and the blue default color

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -74,7 +74,9 @@ pre {
  */
 
 a {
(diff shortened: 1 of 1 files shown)


miquelfire:
What browser does this?

keithort:
@lpeil This project is to normalize the styles across browsers. Color and text decoration on links is an opinionated preference.

clarfonthey:
This is normalize, not reset.

PR #806, opened 2020-01-06
revert the removal of SVG overflow:hidden, still needed for IE

bbenjamin:
The overflow: hidden rule for SVG images was removed in this commit
https://github.com/necolas/normalize.css/commit/004d58b2f2e0ac3d03d075f8de46ce7c8234742
The comment for this commit states:
>   Drop support for older browsers: IE 9-, Android 4, Safari 7-.

This rule is still necessary for supported versions of IE (10/11), so it shouldn't have been removed. This PR brings it back.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -149,6 +149,14 @@ img {
   border-style: none;
 }
 
+/**
+ * Hide the overflow in IE.
+ */
+
+svg:not(:root) {
+  overflow: hidden;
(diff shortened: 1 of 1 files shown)


jakob-e:
Please note that sometimes overflow visible is desirable 
https://twitter.com/JoshWComeau/status/1288606220304158721

PR #817, opened 2020-04-23
HR: no need for Firefox normalization

mattbrundage:
Firefox no longer needs ``box-sizing`` and ``height`` rules, as the browser's stylesheet now specifies the correct ``box-sizing`` value and does not affect height. Tested in Firefox ESR (v. 68) through Firefox Developer Edition (v. 76).

Paste this into Firefox's address bar to view the browser's default styles: ``view-source:resource://gre-resources/html.css``

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -46,14 +46,11 @@ h1 {
    ========================================================================== */
 
 /**
- * 1. Add the correct box sizing in Firefox.
- * 2. Show the overflow in Edge and IE.
+ * 1. Show the overflow in Edge and IE.
  */
 
 hr {
-  box-sizing: content-box; /* 1 */
-  height: 0; /* 1 */
-  overflow: visible; /* 2 */
+  overflow: visible; /* 1 */
 }
 
 /**


SiarheiBobryk:
Completely agree with @mattbrundage!

I can confirm that the `box-sizing` and `height` rules are not required for Firefox anymore. Actually I was going to submit the same changes, but @mattbrundage was first 🙂

According to [User Agent Stylesheet](https://dxr.mozilla.org/mozilla-central/source/layout/style/res/html.css#643) there is already defined `box-sizing: content-box;` style for Firefox.

```css
/* <hr> noshade and color attributes are handled completely by
 * the nsHTMLHRElement attribute mapping code
 */
hr {
  display: block;
  border: 1px inset;
  margin-block-start: 0.5em;
  margin-block-end: 0.5em;
  margin-inline-start: auto;
  margin-inline-end: auto;
  color: gray;
  -moz-float-edge: margin-box;
  box-sizing: content-box;
}

hr[size="1"] {
  border-style: solid none none none;
}
```

Screen Shot 2020-09-13 at 12 06 13 pm

Also, I can confirm that `height` calculates properly in Firefox now.

Screen Shot 2020-09-13 at 12 11 54 pm

SiarheiBobryk:
@necolas up

PR #825, opened 2020-09-06
Fix the input[type="search"] documentation block

SiarheiBobryk:
Changes in README.md:
1. Remove notes about Chrome on macOS and Windows, because it's not
reproduced. See attached images. Tested on macOS 10.15 and Windows 10/7.
2. Remove notes about Safari on Windows, because it's completely
out-of-date and not supported anymore by normalize.css. Safari 5.1.7
was the last version supported by Windows XP/Vista/7.
[Safari version history](https://en.wikipedia.org/wiki/Safari_version_history#Windows)

Changes below are required to cover all cases mentioned on README.md for
Safari.

Changes in test.html:
1. Add some new styles:
    - font-family
    - background-color
2. Add the value attribute

Changes in normalize.css:
1. Fix comment about `-webkit-appearance` for Chrome

Screen Shot 2020-09-06 at 3 16 05 pm

Screen Shot 2020-09-06 at 3 24 29 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,13 +78,10 @@ decrement button to change from `default` to `text`.
 
 #### `[type="search"]`
 
-The search input is not fully stylable by default. In Chrome and Safari on
-OSX/iOS you can't control `font`, `padding`, `border`, or `background`. In
-Chrome and Safari on Windows you can't control `border` properly. It will apply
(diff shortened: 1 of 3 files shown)


SiarheiBobryk:
@necolas up

PR #828, opened 2020-09-15
Remove redundant text-transform for FF's select

SiarheiBobryk:
Nowadays we don't need to specify the text-transform property for the
select element for Firefox. Such functionality already built-in in
Firefox as well as for all other browsers normalise.css supports:
Chrome, Safari, Edge, IE11, IE10.

Sources:
- [Firefox's forms.css](https://dxr.mozilla.org/mozilla-central/source/layout/style/res/forms.css#133)
- `view-source:resource://gre-resources/forms.css`

Other changes:
- Specify IE supported versions because we still need to set
text-transform for button in Edge, IE11, and IE10 explicitly.
- Remove select from the text-transform UI test

Tested on:
- Safari 13.1.2
- Chrome 85.0.4183.102
- Firefox 80.0.1
- Edge 44.19041.423.0
- IE 11.450.19041.0
- IE 10.0.9200.16384

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -179,12 +179,10 @@ input { /* 1 */
 }
 
 /**
- * Remove the inheritance of text transform in Edge, Firefox, and IE.
- * 1. Remove the inheritance of text transform in Firefox.
+ * Remove the inheritance of text transform in Edge, IE11, and IE10.
  */
 
-button,
-select { /* 1 */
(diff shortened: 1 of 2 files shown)


SiarheiBobryk:
Screen Shot 2020-09-15 at 12 18 58 pm

PR #835, opened 2020-10-02
Added background-color normalization

adaamz:
It looks like much websites count with white background as default. But if anybody changes this default color in their browser, the web is completely unusable.
Currently I updated Opera and I have grey background for example on groups.google.com.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -4,13 +4,15 @@
    ========================================================================== */
 
 /**
- * 1. Correct the line height in all browsers.
- * 2. Prevent adjustments of font size after orientation changes in iOS.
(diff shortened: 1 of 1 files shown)


miquelfire:
1. This may be out of the scope of the project, so it may not be accepted. Most sites should be setting that value anyway.
2. If this were to be accepted, however, you should also set the foreground as well. The user might have a setting that sets the background color as a dark color, and the text color as a light color, and this change could make the page unreadable as is.

robsonsobral:
This looks more like https://github.com/csstools/sanitize.css

PR #848, opened 2020-11-24
Fieldsets have common attribute and size is widen than I think..

oinochoe:
I had in trouble with code without attribute 'min-inline-size:auto'.. 
Because many attributes have min-inline-size..
Especially, fieldsets have that size, and I can't make this element to 'width: 100%'.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -228,6 +228,7 @@ button:-moz-focusring,
 
 fieldset {
   padding: 0.35em 0.75em 0.625em;
+  min-inline-size:auto;
 }
 
 /**


mattbrundage:
It's not appropriate to override `min-inline-size` for ``fieldset``. Both Chromium and Gecko apply `min-inline-size: min-content` to fieldsets in their UA stylesheets, while Safari applies a similar `min-width: min-content`.  So `min-content` looks like a clear precedent.

oinochoe:
> It's not appropriate to override `min-inline-size` for `fieldset`. Both Chromium and Gecko apply `min-inline-size: min-content` to fieldsets in their UA stylesheets, while Safari applies a similar `min-width: min-content`. So `min-content` looks like a clear precedent.

@mattbrundage exactly, i think browser supporters will make it better solution, and they have to to that.
more clearly, we can still overwrite the attribute. thanks for the reply : )

Issue #851, opened 2021-01-24
Word-break inconsistency between Chrome and Firefox


Gaboris:
While my friend was working on a project to learn how to build a webpage they've found that Chrome doesn't wrap its links at all while Firefox does.

After taking a look at the matter it seems that Chrome uses a different default value for `word-break` then Firefox as after adding `word-break: break-word;` on its links it behaved the same way as Firefox did.

Can anyone confirm this? I can't remember where to find the default styles right no, but it would be nice to follow FF's more reasonable approach.

I also saw that similar questions have come up before and they were put aside because of "unexpected behaviors", but this seems like a special case as it to only happens with the `<a>` tag so it could be an exception.

My friend also said that they'll do a fork and add a fix for it in case the project would want to add it in as a pull request.

VixieTSQ:
I'm said friend. Just created the pull request. I tested it with another friend it works with their Chrome.

PR #852, opened 2021-01-24
Correct the link word wrapping within a container in Chrome.

VixieTSQ:
To do with #851.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -70,11 +70,13 @@ pre {
    ========================================================================== */
 
 /**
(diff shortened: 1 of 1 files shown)


VixieTSQ:
This will never get merged. I assume this repo to be dead.

garrettw:
Considering adding this to my normalize in https://github.com/garrettw/vitals -- but only if your code matches what most browsers tend to do (i.e. needs to be more than just Firefox). If most browsers do what Chrome does, then the fix should correct Firefox's behavior.

Issue #860, opened 2021-06-28
hr normalize applying wrong overflow value


brady-shinners-bby:
[HTML5 spec for hr is to apply `overflow: hidden`](https://html.spec.whatwg.org/multipage/rendering.html#the-hr-element-2), but [normalize is currently applying `overflow: visible`](https://github.com/necolas/normalize.css/blob/master/normalize.css#L56). Please either remove what appears to be an unneeded override of a now-uniform default style or apply the expected overflow: hidden.

garrettw:
Are you confirming that most browsers are respecting the spec now, or that the override definitely isn't needed anymore?

brady-shinners-bby:
The override does the opposite of the spec. I'm not familiar with what normalize's target browsers are, so not sure on whether overflow: visible should be replaced with overflow: hidden or just removed. For my use-case, all browsers we support adhere to overflow: hidden.

garrettw:
In my normalize in https://github.com/garrettw/vitals I'm making the change to `hidden` until I know for sure that it isn't needed anymore.

PR #861, opened 2021-07-15
Changed overflow: visible to overflow: hidden for hr element

costingh:
Applied the expected overflow value for hr element, accordingly to HTML5 spec

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -53,7 +53,7 @@ h1 {
 hr {
   box-sizing: content-box; /* 1 */
   height: 0; /* 1 */
-  overflow: visible; /* 2 */
+  overflow: hidden; /* 2 */
 }
 
 /**
--- a/normalize.css
+++ b/normalize.css
@@ -47,7 +47,7 @@ h1 {
 
 /**
  * 1. Add the correct box sizing in Firefox.
- * 2. Show the overflow in Edge and IE.
+ * 2. Hide the overflow in Edge and IE.
  */
 
 hr {


normalize.css
```diff
@@ -53,7 +53,7 @@ h1 {
 hr {
   box-sizing: content-box; /* 1 */
   height: 0; /* 1 */
-  overflow: visible; /* 2 */
+  overflow: hidden; /* 2 */
```
nosilleg, line 56: With this change the comment on line number 50 is now incorrect.

nuxodin:
Do you have a link to the spec?
I can't find anything regarding css overflow.

costingh:
> Do you have a link to the spec?
> I can't find anything regarding css overflow.

This is the link to the spec: https://html.spec.whatwg.org/multipage/rendering.html#the-hr-element-2

normalize.css
costingh: you are right, I need to look into more detail

costingh: Fixed the comment. Hope it is ok now :)

nosilleg: Looking at the comment it would appear that Edge and IE were following spec and all other browsers weren't. So your change to the comment is still not accurate.

I'm not a maintainer so I don't know what the full expectations are for this, but for me personally I would expect the comment to reference the browsers that show overflow. This requires testing. If there are no browsers that are still showing overflow then the code change should be to remove the lines, not to change the value to "hidden". If there are still browsers showing overflow then it would be good to have that list so that eventually the code can be removed from this repo when they all follow spec.

nosilleg:
fixes #860

normalize.css
garrettw: Adding this to my normalize in https://github.com/garrettw/vitals until I know it's not needed anymore.

PR #876, opened 2022-02-19
Remove duplicate font name

jdromero88:


nosilleg:
If you believe this is no longer required then please update or remove the comment as well as the code change. Please also provide the reasoning you believe it's no longer required.

The current state of this PR gives the impression that you haven't read the comment and understood why it exists.

PR #879, opened 2022-03-24
feature: If `hidden="until-found"`, it's not `display: none`.

yamanoku:
`hidden` attribute, which used to have only a boolean value, now also has a new `until-found` value.

ref: https://github.com/whatwg/html/commit/f5def65bda4d01a6a1e105fcf13fe45201b4f48a

Elements with `hidden="until-found"` will have `content-visibility: hidden` applied instead of `display: none`.
If specified with `display: none`, it may not be found in in-page searches.

ref: [HTML の hidden 属性が列挙型に変更され hidden="until-found" が追加](https://blog.w0s.jp/664)

To avoid this problem, the style specification for `hidden` case has been modified.

sample: https://codepen.io/yamanoku/pen/YzYpoqG

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -342,8 +342,9 @@ template {
 
 /**
  * Add the correct display in IE 10.
+ * Except `hidden="until-found"`　because `content-visibility: hidden` is applied.
  */
 
-[hidden] {
+[hidden]:not([hidden="until-found"]) {
   display: none;
 }
--- a/normalize.css
+++ b/normalize.css
@@ -345,6 +345,10 @@ template {
  * Except `hidden="until-found"`　because `content-visibility: hidden` is applied.
  */
 
-[hidden]:not([hidden="until-found"]) {
+[hidden] {
   display: none;
+}
+
+[hidden="until-found"] {
+  display: revert;
 }
--- a/test.html
+++ b/test.html
@@ -142,6 +142,7 @@
       <content></content>
     </template>
     <p hidden>This should be hidden</p>
+    <p hidden="until-found">This is not display none</p>
   </div>
 
   <h2 class="Test-describe"><code>a</code></h2>


normalize.css
```diff
@@ -342,8 +342,9 @@ template {
 
 /**
  * Add the correct display in IE 10.
+ * Except `hidden="until-found"`　because `content-visibility: hidden` is applied.
```
thibaudcolas, line 345:
```suggestion
 * Add the correct display in IE 10, and for SVG elements.
 * Except `hidden="until-found"`　because `content-visibility: hidden` is applied.
```

thibaudcolas (commented):
I came here looking exactly for this change, great to see this is already being worked on!

Just wanted to also suggest updating the comment to make it clear this isn’t just for legacy browser support – `hidden` also [isn’t supported by SVG elements](https://github.com/w3c/svgwg/issues/508) inline in HTML, which is very surprising. I’m not sure if this was considered when creating this specific bit of the reset, but now that this code is widely used it’d be problematic to remove this for projects that don’t care about IE10 support. 

See for example [Tailwind discussion on a similar change](https://github.com/tailwindlabs/tailwindcss/discussions/8247#discussioncomment-2919589).

normalize.css
```diff
@@ -342,8 +342,9 @@ template {
...
  */
 
-[hidden] {
+[hidden]:not([hidden="until-found"]) {
   display: none;
 }
```
thibaudcolas: `[hidden]:not([hidden="until-found"])` has [higher specificity](https://polypane.app/css-specificity-calculator/#selector=%5Bhidden%5D%3Anot(%5Bhidden%3D%22until-found%22%5D)) than the previous selector due to the `:not`. I wonder if this should instead be written as two separate declarations, so there is no increase in specificity? This will be less likely to cause issues for existing projects relying on `[hidden]` having a very low specificity.

```suggestion
[hidden] {
  display: none;
}

[hidden="until-found"] {
  display: revert;
}
```

normalize.css
yamanoku: > `hidden` also [isn’t supported by SVG elements](https://github.com/w3c/svgwg/issues/508 ) inline in HTML,

Oh. I did not know this fact. Thank you !

huajie2020 (approved):

huajie2020 (approved):
dunmo

huajie2020 (commented):
- [ ]

normalize.css
```diff
@@ -342,8 +342,13 @@ template {
...
 /**
  * Add the correct display in IE 10.
+ * Except `hidden="until-found"`　because `content-visibility: hidden` is applied.
  */
 
 [hidden] {
   display: none;
 }
+
+[hidden="until-found"] {
```
abdellah259, line 352: good

PR #881, opened 2022-06-05
Default list style position to "outside"

LexNastin:
As per MDN list-style-position should default to outside, but after some testing, every Chromium based browser defaults this to inside. In IE 11 and Firefox 101, it defaults to outside, as it should.

**Edit:** Safari apparently also defaults this to inside. This quirk is [mentioned in MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position) right before the "Syntax" section on the page.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -347,3 +347,11 @@ template {
 [hidden] {
   display: none;
 }
+
+/**
+ * Default list style to "outside" in Chrome
+ */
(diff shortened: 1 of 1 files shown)


garrettw:
Adding this to my normalize in https://github.com/garrettw/vitals

PR #894, opened 2023-01-20
correct summary styling

### LINKED ISSUE #891: summary element styling needs to be scoped to details element
The current display style applied to the summary element results in unwanted list-item styling for instances where the summary element is not used properly.  

```
summary {
  display: list-item;
}
```

This should rather be something like `details > summary:first-of-type` - which matches how modern browsers are treating the element.

scottaohara:
closes #891

This revised style ensures that only summary elements that are properly used will be provided the `display: list-item` style.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -325,7 +325,7 @@ details {
  * Add the correct display in all browsers.
  */
 
-summary {
+details > summary:first-of-type {
   display: list-item;
 }
 


garrettw:
Adding this to my normalize in https://github.com/garrettw/vitals

samuelreichor:
I would suggest du completely hide the default detail marker like that:

```css
summary::-webkit-details-marker {
 display: none;
}

summary {
 list-style: none;
}
```

PR #903, opened 2023-07-24
se agrego max-width: 100%; a la img

perch733:
I have inserted
img{
max-width:100%;
} to make the image responsive and not crash.

### DIFF
--- a/normalize.css
+++ b/normalize.css
@@ -147,6 +147,7 @@ sup {
 
 img {
   border-style: none;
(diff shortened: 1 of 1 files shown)


hachesilva:
Hola @perch33,

Not trying to steal the mic from the project maintainers but I think this might be closed sooner or later:
As explained in other PRs, not all images need a 100% max width, see here for example https://github.com/necolas/normalize.css/pull/166

Issue #929, opened 2025-07-01
change the UI of this page


PRITAM-TU:
change the UI of this page
![Image](https://github.com/user-attachments/assets/4604227f-65e6-4416-a663-8a53a7825a8c)

garrettw:
This is not helpful. Please close this issue.

PR #930, opened 2025-07-01
add UI in css file

PRITAM-TU:
I was add only UI section if that is good for you then later i will contribute :
fixes:#929
change the UI of this page

Uploading Recording 2025-07-01 114204.mp4…


garrettw:
This is not helpful at all. Please close this PR.

